Issue #11, opened 2020-03-29
Per index stats


JokerQyou:
It seems like the API for per-index stats information is currently not available with this client library, is that right? `client.Stats().Get()` always returns an empty stats data.

alexisvisco:
Thanks you for your report, I will investigate this soon !

alexisvisco:
There was indeed a problem. The client was using the wrong route.

Issue #7, opened 2020-03-25
AttributesToHighlight example does not work


curquiza:
When I try the example in the readme with the `AttributesToHighlight` parameter:

```go
resp, err := client.Search(indexUID).Search(meilisearch.SearchRequest{
    Query: "harry pottre",
    AttributesToHighlight: "*",
})
```

it does not run, with this error:

```bash
# command-line-arguments
./test.go:47:3: cannot use "*" (type string) as type []string in field value
```

Issue #4, opened 2020-03-24
Search parameters should be optional


curquiza:
The search returns no result if I don't pass the `Limit` parameter. It does not fail. It just returns empty `Hits`.

```go
_, err := client.Search("indexUID").Search(meilisearch.SearchRequest{
	Query: "harry pottre",
})
```
=> no result

```go
_, err := client.Search("indexUID").Search(meilisearch.SearchRequest{
	Query: "harry pottre",
        Limit: 10,
})
```
=> result

We should be able to pass only the `Query` parameter

### Edit

<s>I understand that 2 parameters at least are needed because
```go
searchRes, err := client.Search("indexUID").Search(meilisearch.SearchRequest{
	Query: "harry pottre",
	AttributesToHighlight: "*",
})
```
returns results as well.</s>

My bad, it does not even run at all, I open an issue for that: #7 

### Version
`v0.9.0` of MeiliSearch.

alexisvisco:
Hi thanks for your report: The default value of an int in Go is 0 so if you don't initialize the field `Limit` from the `SearchRequest` you get 0 result: it's normal.

Maybe we can provide a default limit in accordance to the MeiliSearch engine ?

curquiza:
Hi!

The MeiliSearch engine uses a limit of 20 by default.

But isn't it possible to provide a function that sends a simple request without any parameter if the user does not provide any? 🙂 Except for the mandatory one (`query`) of course

alexisvisco:
With the fix added in  `dev` it should be possible :)

PR #12, opened 2020-03-30
Fix: stats client, default limit value in search, example in readme

alexisvisco:
The stats client was incomplete and wrong:
It was using the `/stats` endpoint to get the stats for an index. 

The stats client now has two routes:
- Get
- GetAll

Resolve issue #11


alexisvisco:
Fixing default value for `limit` parameter in search request. #4

alexisvisco:
Fixing readme example: a list of string is not a string. #7

qdequele:
Is the PR ready for review?

alexisvisco:
Yes it is

client.go
```diff
@@ -229,34 +230,37 @@ func (c Client) handleResponse(req *internalRequest, response *http.Response, in
...
+// WaitForPendingUpdate waits for the end of an update.
+// The function will check by regular interval provided in parameter interval
+// the UpdateStatus. If it is not UpdateStatusEnqueued or the ctx cancelled
+// we return the UpdateStatus.
+func (c Client) WaitForPendingUpdate(
+	ctx context.Context,
+	interval time.Duration,
+	indexID string,
+	updateID *AsyncUpdateID) (UpdateStatus, error) {
+
```
qdequele, line 251: I don't find it natural not to be able to use this method only with context. What do you think about just creating the `WaitForPendingUpdate` with the content of the `defaultWaitForPendingUpdate `. We can take a radical choice and force the interval to 1 second and the timeout to 1 minute. In the future, we will probably add a channel to receive updates status. 
What do you think @alexisvisco @curquiza @eskombro?

curquiza: I'm not really comfortable at all with the idea of providing a function with a timeout that the user can not manipulate. Not everyone accepts to wait for 1 minute for example. This kind of function **must** be customizable for me.

But if you want guys, we can pass the function `defaultWaitForPendingUpdate` in public again? In this case, sorry @alexisvisco for the useless change in the previous PR 😅
The user would have 2 choices: the default one with `defaultWaitForPendingUpdate`, and the customizable one with `WaitForPendingUpdate`.

curquiza: (Plus, that's weird that this change is in this PR, because I merged it in the previous one. Can you rebase please? 🙂)

eskombro: I agree, it is important to be able to customize a timeout. Could it be a field in the Client config struct? So we have a default value for a client, no need for params, but can be modified by user?

curquiza: We thought about this default `timeout` value with Charlotte too 👍 
It might come with the next big changes, but not a priority for the moment. We'll apply that in all the SDK in the same time :)

curquiza: (Except if someone has time to do it of course haha)

eskombro: Then it might be an option, as you said, to export both functions (`defaultWaitForPendingUpdate` => `DefaultWaitForPendingUpdate`) while some customization options/methods are actually integrated

alexisvisco: If we provide a way to define the interval and the timeout in the client this will confuse the user: 

- he may think timeout is for all http requests
- he may think why the fuck there is an interval option in the client while this parameter is used only for one function.

I will say again what I said in the pr: the user can create his own function with default parameters if he want.

If the library provide default values for the `WaitForPendingUpdate` then it will be necessary to experiment to take the best values according some benchmarks. It may be a waste of time to worry about it.

alexisvisco: Could possibly solve all this questions:

```go
client.WaitForPendingUpdate("indexId", updateID, withWaithForPendingUpdateContext(ctx), withWaitForPendingUpdateInterval(1 * time.seconds))
```

Functional option parameter is widely used in go and is a good practice. Here if the user does not provide withWaithForPendingUpdateContext or withWaitForPendingUpdateInterval: default values will be set. 

Why this painful naming: withWaithForPendingUpdate... ? 
Because if the client take function option parameter in the future I prefer let the name `with` + `the name of the option` to the client.

curquiza: > he may think why the fuck there is an interval option in the client while this parameter is used only for one function.

Don't worry, it's only about the timeout in the client, not the interval 😉
But indeed, the "general" timeout (defined in the client) is not necessary the time to wait for the update but more for the HTTP requests. So it makes no sense to define the timeout for the update in the client, right.

> Functional option parameter is widely used in go and is a good practice. Here if the user does not provide withWaithForPendingUpdateContext or withWaitForPendingUpdateInterval: default values will be set.

I'm glad to see there is a possibility to call a function without some parameters in go! That's what I'm asking since the beginning!! 😂
So, I can use the function with this syntax: `client.WaitForPendingUpdate("indexUID", updateID)` without specify the two last parameters, and you confirm it will work with the default values? @alexisvisco

alexisvisco: Yes you can: https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis

qdequele: Hi, sorry for the confusion. We've had a long talk with the team and we like the basic idea of having a `WaitForPendingUpdate` which is configurable and a `DefaultWaitForPendingUpdate` with default configuration as you write before. I'm sorry again for all the changes. If you want to talk about it in real-time, I would like to invite you to our [slack community](slack.meilisearch.com) 😃.

alexisvisco: @qdequele You want the Default method to be public ?

qdequele: Yes, please 😀. And again, thanks a lot for your work!

qdequele:
@eskombro can you review the code?

qdequele:
@eskombro can you review the code?

qdequele:
@eskombro can you review the code?

Issue #10, opened 2020-03-25
Infinite loop in AwaitAsyncUpdateID function


curquiza:
If I have well understood, we are stuck in an infinite loop, if for any reason, the status of the update is stuck to `enqueued`.

https://github.com/meilisearch/meilisearch-go/blob/f9427a2cc1e502de018a355ac65e4c81262e703e/client.go#L241-L256

We should define a default value of timeout.
This value could be change when passing an optional `timeOut` parameter in this function.

Plus, the name of the `updateID` parameter is not really accurate: the function expects an update "object" and not a updateID.

alexisvisco:
Yes we can add a ctx for sure that let user define a timeout :)
That's a nice idea.

PR #13, opened 2020-03-30
Improvement: Method AwaitAsyncUpdateID

alexisvisco:
The method has been changed to add two things:
- A context to manipulate from the outside when the function should stop polling the status of an update id.
- A interval duration to configure the time between each poll. 

A simplified method called `AwaitAsyncUpdateIDSimplified` set default to `AwaitAsyncUpdateID` with a timeout of `5 seconds` and a interval of `50ms`. 

Resolve issue #10

### DIFF
--- a/client.go
+++ b/client.go
@@ -239,20 +239,20 @@ func contains(slice []int, val int) bool {
 	return false
 }
 
-// AwaitAsyncUpdateID check each 50ms the status of a AsyncUpdateID.
-// This method should be avoided.
-func (c Client) AwaitAsyncUpdateIDSimplified(indexUID string, updateID *AsyncUpdateID) (UpdateStatus, error) {
+// defaultWaitForPendingUpdate check each 50ms the status of a AsyncUpdateID.
+// This method is used for test purpose
+func (c Client) defaultWaitForPendingUpdate(indexUID string, updateID *AsyncUpdateID) (UpdateStatus, error) {
 	ctx, cancelFunc := context.WithTimeout(context.Background(), time.Second*5)
 	defer cancelFunc()
 
-	return c.AwaitAsyncUpdateID(ctx, time.Millisecond*50, indexUID, updateID)
+	return c.WaitForPendingUpdate(ctx, time.Millisecond*50, indexUID, updateID)
 }
 
-// AwaitAsyncUpdateID wait for the end of an update.
+// WaitForPendingUpdate wait for the end of an update.
 // The function will check by regular interval provided in parameter interval
 // the UpdateStatus. If it is not UpdateStatusEnqueued or the ctx cancelled
 // we return the UpdateStatus.
-func (c Client) AwaitAsyncUpdateID(
+func (c Client) WaitForPendingUpdate(
 	ctx context.Context,
 	interval time.Duration,
 	indexID string,
--- a/client.go
+++ b/client.go
@@ -239,7 +239,7 @@ func (c Client) defaultWaitForPendingUpdate(indexUID string, updateID *AsyncUpda
 	return c.WaitForPendingUpdate(ctx, time.Millisecond*50, indexUID, updateID)
 }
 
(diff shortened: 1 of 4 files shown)


curquiza:
Hello! Sorry for the delay!

Thanks for your PR! And for your time answering the issues! It's really helpful!

Could you change some points?
- The name of the function `AwaitAsyncUpdateID`: we talked about that with the team and we would rather use `waitForPendingUpdate` which is more accurate 🙂
- The function `AwaitAsyncUpdateIDSimplified` is more a tool for the tests than a real function we want to provide (I even see that you write "This method should be avoided"). Could you put it in a kind of `utils` file that you import in all the tests instead of providing it in the Client?

alexisvisco:
So I have made AwaitAsyncUpdateIDSimplified private. I think ifan  user want this method with a default behavior he can  totally doing this by himself.

client.go
```diff
@@ -229,34 +230,37 @@ func (c Client) handleResponse(req *internalRequest, response *http.Response, in
...
-func contains(slice []int, val int) bool {
-	for _, item := range slice {
-		if item == val {
-			return true
-		}
-	}
-	return false
+// defaultWaitForPendingUpdate check each 50ms the status of a AsyncUpdateID.
+// This method is used for test purpose
+func (c Client) defaultWaitForPendingUpdate(indexUID string, updateID *AsyncUpdateID) (UpdateStatus, error) {
```
curquiza, line 235: Since this method is for tests purpose and has nothing to do with the client, can we put it in another file like `test_utils.go` for example? 🙂 Or another name, I don't know the convention in go.

client.go
```diff
@@ -229,34 +230,37 @@ func (c Client) handleResponse(req *internalRequest, response *http.Response, in
...
+	defer cancelFunc()
+
+	return c.WaitForPendingUpdate(ctx, time.Millisecond*50, indexUID, updateID)
 }
 
-// AwaitAsyncUpdateID check each 16ms the status of a AsyncUpdateID.
-// This method should be avoided.
-// TODO: improve this method by returning a channel
-func (c Client) AwaitAsyncUpdateID(indexID string, updateID *AsyncUpdateID) UpdateStatus {
+// WaitForPendingUpdate wait for the end of an update.
```
curquiza, line 242:
```suggestion
// WaitForPendingUpdate waits for the end of an update.
```

client.go
curquiza: My bad, I haven't seen your comment about making it private. I would rather put it in another file because this method is for our only tests.

But If it's a good practice in golang, so, let's do that.

curquiza: @qdequele @eskombro can you confirm?

alexisvisco: It's more common to have all methods from a struct in the file where the struct is declared. 
But I can refactor it to take the client in param and move it to a test_utils.go file if you want. It's the same ^^

qdequele: I didn't know the good practices about that sorry 😏

curquiza: Ok, let's keep your choice then @alexisvisco.
But the last question: if it it's a private function, how is it possible to call it?

client.go
```diff
@@ -229,34 +230,37 @@ func (c Client) handleResponse(req *internalRequest, response *http.Response, in
...
 }
 
-func contains(slice []int, val int) bool {
-	for _, item := range slice {
-		if item == val {
-			return true
-		}
-	}
-	return false
+// defaultWaitForPendingUpdate check each 50ms the status of a AsyncUpdateID.
```
curquiza, line 233:
```suggestion
// defaultWaitForPendingUpdate checks each 50ms the status of a WaitForPendingUpdate.
```

client.go
curquiza: I mean, to call it in our tests?

alexisvisco: All tests are in the same package :)

curquiza: Ok good to know that it works in this case :) thx!

PR #21, opened 2020-04-29
Update go.yml

curquiza:
Run the tests on pull requests (does not work on the PR so far)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -1,5 +1,5 @@
 name: Go
-on: [push]
+on: [pull_request]
 jobs:
 
   build:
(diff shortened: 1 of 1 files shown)


curquiza:
I added a rule to avoid merging the branch if it's not up-to-date with `master`, so it will only duplicate the tests: they will run twice with exactly the same context 😉
Only administrators can overpass this rule, and we are human, so we can do mistakes. I can add it if you worry about that!

curquiza:
I'm going to close this PR, I did tests on it with the CI. I realized the behavior of the current CI is what we expect 🙂 (I thought it was not because the tests did not run on the previous PR). Sorry guys!

PR #31, opened 2020-05-18
Fix 🐛 Documents.List()

derekahn:
## Problem
It always returned all documents regardless of query params set

## Solution
perform request with the required query params vs as a payload

### DIFF
--- a/client_documents_test.go
+++ b/client_documents_test.go
@@ -143,11 +143,9 @@ func TestClientDocuments_List(t *testing.T) {
 		Host: "http://localhost:7700",
 	})
 
-	_, err := client.Indexes().Create(CreateIndexRequest{
+	if _, err := client.Indexes().Create(CreateIndexRequest{
 		UID: indexUID,
-	})
-
-	if err != nil {
+	}); err != nil {
 		t.Fatal(err)
 	}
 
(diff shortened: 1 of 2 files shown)


curquiza:
Hello @derekahn! Thanks for noticing this issue and fixing it! 😁
I've just requested the author of this package and a Go developer of the Meili team.

client_documents_test.go
```diff
@@ -166,17 +164,16 @@ func TestClientDocuments_List(t *testing.T) {
...
 	}, &list)
 
 	if err != nil {
 		t.Fatal(err)
 	}
 
-	// tests are running in parallel so there can be more than 1 docs
-	if len(list) < 1 {
-		t.Fatal("number of doc should be at least 1")
+	if list[0].ID != "456" {
```
eskombro, line 175:
```suggestion
	if len(list) == 0 || list[0].ID != "456" {
```

eskombro: This test actually replaced the old one in your PR, which was quite simple but still seems pertinent. Also, an error in retrieving the list would cause a crash instead of just failing the test. So I suggest adding this simple check before trying to access the slice first element, inb case it doesn't exist. We have the double check and avoid a crash

derekahn: Good idea including that safety in the test 👍.

eskombro:
@alexisvisco I merged this PR because I wanted to fix the issue fast enough, but if you have any comment feedback don't hesitate to come back over it! 👍

derekahn:
Thank you guys for all of your hard work on this for people like myself can use ❤️.

PR #39, opened 2020-06-08
Implement faceting

eskombro:
- [x] Implement faceting setting routes
- [x] Implement facetsDistribution search parameter
- [x] Implement facetFilters search parameter
- [x] Improve search tests (add test for existing params and add facetsDistribution and facetFilters)
- [x] add faceting routes tests

For obvious reasons, tests won't pass as faceting is not compatible with the current latest release (v0.10.1)

### DIFF
--- a/client_documents_test.go
+++ b/client_documents_test.go
@@ -10,9 +10,9 @@ type docTest struct {
 }
 
 type docTestBooks struct {
-	Book_id int    `json:"book_id"`
-	Title   string `json:"title"`
-	Tag     string `json:"tag"`
+	BookId int    `json:"book_id"`
+	Title  string `json:"title"`
+	Tag    string `json:"tag"`
 }
 
 func TestClientDocuments_Get(t *testing.T) {
--- a/client_documents_test.go
+++ b/client_documents_test.go
@@ -9,6 +9,12 @@ type docTest struct {
 	Name string `json:"name"`
 }
 
+type docTestBooks struct {
+	Book_id int    `json:"book_id"`
+	Title   string `json:"title"`
+	Tag     string `json:"tag"`
+}
+
 func TestClientDocuments_Get(t *testing.T) {
 	var indexUID = "TestClientDocuments_Get"
 
--- a/client_documents_test.go
+++ b/client_documents_test.go
@@ -10,7 +10,7 @@ type docTest struct {
 }
 
 type docTestBooks struct {
-	BookId int    `json:"book_id"`
+	BookID int    `json:"book_id"`
 	Title  string `json:"title"`
 	Tag    string `json:"tag"`
 }
--- a/client_search.go
+++ b/client_search.go
@@ -1,6 +1,7 @@
 package meilisearch
 
 import (
+	"fmt"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -50,6 +51,9 @@ func (c clientSearch) Search(request SearchRequest) (*SearchResponse, error) {
 	}
 	if request.Matches {
 		values.Add("matches", strconv.FormatBool(request.Matches))
+	}
+	if len(request.FacetsDistribution) != 0 {
+		values.Add("facetsDistribution", fmt.Sprintf("[\"%s\"]", strings.Join(request.FacetsDistribution, "\",\"")))
 	}
 
 	req := internalRequest{
--- a/client_search.go
+++ b/client_search.go
@@ -53,7 +53,7 @@ func (c clientSearch) Search(request SearchRequest) (*SearchResponse, error) {
 		values.Add("matches", strconv.FormatBool(request.Matches))
 	}
 	if len(request.FacetsDistribution) != 0 {
-		values.Add("facetsDistribution", fmt.Sprintf("[\"%s\"]", strings.Join(request.FacetsDistribution, "\",\"")))
+		values.Add("facetsDistribution", fmt.Sprintf("[%q]", strings.Join(request.FacetsDistribution, "\",\"")))
 	}
 	if request.FacetFilters != nil {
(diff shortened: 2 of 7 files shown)


types.go
```diff
@@ -20,13 +20,14 @@ type Index struct {
 
 // Settings is the type that represent the settings in MeiliSearch
```
curquiza, line 21:
```suggestion
// Settings is the type that represents the settings in MeiliSearch
```

types.go
```diff
@@ -20,13 +20,14 @@ type Index struct {
...
+	DistinctAttribute     *string             `json:"distinctAttribute,omitempty"`
+	SearchableAttributes  []string            `json:"searchableAttributes,omitempty"`
+	DisplayedAttributes   []string            `json:"displayedAttributes,omitempty"`
+	StopWords             []string            `json:"stopWords,omitempty"`
+	Synonyms              map[string][]string `json:"synonyms,omitempty"`
+	AcceptNewFields       bool                `json:"acceptNewFields,omitempty"`
+	AttributesForFaceting []string            `json:"attributesForFaceting,omitempty"`
 }
 
 // Version is the type that represent the versions in MeiliSearch
```
curquiza, line 33:
```suggestion
// Version is the type that represents the versions in MeiliSearch
```

types.go
```diff
@@ -20,13 +20,14 @@ type Index struct {
...
 // Settings is the type that represent the settings in MeiliSearch
 type Settings struct {
-	RankingRules         []string            `json:"rankingRules,omitempty"`
-	DistinctAttribute    *string             `json:"distinctAttribute,omitempty"`
-	SearchableAttributes []string            `json:"searchableAttributes,omitempty"`
-	DisplayedAttributes  []string            `json:"displayedAttributes,omitempty"`
-	StopWords            []string            `json:"stopWords,omitempty"`
-	Synonyms             map[string][]string `json:"synonyms,omitempty"`
-	AcceptNewFields      bool                `json:"acceptNewFields,omitempty"`
+	RankingRules          []string            `json:"rankingRules,omitempty"`
```
curquiza, line 23: Does `omitempty` mean that it can be empty? 🙂

eskombro: In go every variable is automatically initialized to a default zero value (an int is 0, a string is "", a pointer will be nil, etc...). This means that when you create a struc, every field is initialized. For example:

```go
type struct Person {
    Size     int   `json:"size,omitempty"`
    Age      int   `json:"age,omitempty"`
}

```

and you create a Person object not defining on element (Age for example) like this:

```go
person := Preson { Size: 170 }
```

it will be initialized with the default zero values for each property (it would be the same as saying Person { Size: 170, Age 0 }

client_documents_test.go
```diff
@@ -9,6 +9,12 @@ type docTest struct {
 	Name string `json:"name"`
 }
 
+type docTestBooks struct {
+	Book_id int    `json:"book_id"`
```
alexisvisco, line 13: Use  go convention, should be BookID.

client_search.go
```diff
@@ -51,6 +52,13 @@ func (c clientSearch) Search(request SearchRequest) (*SearchResponse, error) {
 	if request.Matches {
 		values.Add("matches", strconv.FormatBool(request.Matches))
 	}
+	if len(request.FacetsDistribution) != 0 {
+		values.Add("facetsDistribution", fmt.Sprintf("[\"%s\"]", strings.Join(request.FacetsDistribution, "\",\"")))
```
alexisvisco, line 56: Use %q to format a string with quote.

client_search.go
```diff
@@ -76,3 +84,27 @@ func (c clientSearch) IndexID() string {
...
 	return c.client
 }
+
+func facetFiltersToStr(i interface{}) string {
+	facetsToStr := ""
+	switch v := i.(type) {
+	case []string:
+		for _, slice := range v {
+			stringSlice := slice
+			facetsToStr += fmt.Sprintf("\"%s\",", stringSlice)
```
alexisvisco, line 94: Same as above, use %q, btw if you can have a quote in your string check if the %q escape the quotes, if it's not the case maybe it will be judicious to escape them.

client_search.go
```diff
@@ -76,3 +84,27 @@ func (c clientSearch) IndexID() string {
...
+		}
+		facetsToStr = fmt.Sprintf("[%s]", facetsToStr[:len(facetsToStr)-1])
+	case [][]string:
+		for _, mainSlice := range v {
+			facetsToStr += "["
+			for _, slice := range mainSlice {
+				stringSlice := slice
+				facetsToStr += fmt.Sprintf("\"%s\",", stringSlice)
```
alexisvisco, line 102: Same as above, use %q.

client_search.go
alexisvisco: Yes it escape all quotes: https://play.golang.org/p/sh-qd4kSPTO

eskombro: Great, didn't know that!

client_documents_test.go
eskombro: Indeed, thanks! Got mixed with the documents syntax :)

eskombro:
Thanks for your review @alexisvisco ! I made the changes you proposed!

client_documents_test.go
```diff
@@ -9,6 +9,12 @@ type docTest struct {
 	Name string `json:"name"`
 }
 
+type docTestBooks struct {
+	BookId int    `json:"book_id"`
```
alexisvisco, line 13: Ahhhh So close to the goal :laughing:  
BookdID not BookId (in go an acronym is always in uppercase)

types.go
curquiza: Thansk!!!

PR #42, opened 2020-06-11
Implement deleteAllIndexes() in tests, use before and after running tests

### LINKED ISSUE #8: Delete indexes after tests
Currently, if I have already a MeiliSearch isntance running locally, and if I run the tests `./run_tests.sh`, there are several indexes still running in my MeiliSearch instance.

We should provide tests that clean all the indexes created for tests purpose.

eskombro:
- [x] Implement deleteAllIndexes() 
- [x] Use deleteAllIndexes() before and after tests

Closes #8

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -1,0 +1,75 @@
+package meilisearch
+
+import (
+	"os"
+	"testing"
+)
+
+type docTest struct {
+	ID   string `json:"id"`
+	Name string `json:"name"`
+}
+
+type docTestBooks struct {
+	BookID int    `json:"book_id"`
+	Title  string `json:"title"`
+	Tag    string `json:"tag"`
+}
+
+func deleteAllIndexes(client *Client) (ok bool, err error) {
+	list, err := client.Indexes().List()
+
+	if err != nil {
+		return false, err
+	}
+
+	for _, index := range list {
+		client.Indexes().Delete(index.UID)
+	}
+
+	return true, nil
+}
+
+func TestMain(m *testing.M) {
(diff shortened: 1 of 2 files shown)


main_test.go
```diff
@@ -0,0 +1,75 @@
...
+	var client = NewClient(Config{
+		Host: "http://localhost:7700",
+	})
+	deleteAllIndexes(client)
+	m.Run()
+	deleteAllIndexes(client)
+	os.Exit(0)
+}
+
+func TestClientIndexes_deleteAllIndexes(t *testing.T) {
```
curquiza, line 43: Is it a test for `deleteAllIndexes` function?

main_test.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	for _, index := range list {
+		client.Indexes().Delete(index.UID)
+	}
+
+	return true, nil
+}
+
+func TestMain(m *testing.M) {
```
curquiza, line 33: Where is this function used?

main_test.go
eskombro: This used to be in a different file, name is wrong, will fix now!

main_test.go
eskombro: Every time you run `go test` for 'meilisearch' module, it will run `TestMain` which will delete all indexes, run the test as usual and finally delete all indexes again.

So basically, it is triggered by `go test`, `go test -v`, `go test -v ./...`

curquiza: Oh magic!!

Issue #43, opened 2020-06-11
Use URL query parameter builder 


alexisvisco:
[This code ](https://github.com/meilisearch/meilisearch-go/blob/b6a24d611fe4ca61813d68b44792db6778105410/client.go#L173) is unsafe for the developer who is using the library.

Query parameters should be built with the URL type. (https://stackoverflow.com/a/30657518/5894301)

**Why is it unsafe ?**

Imagine the developer who is using the APIDocuments.List set AttributesToRetrieve to an input from an user of his website. A malicious user can give an input like that: ?limit=300000.

**What could be cool ?**

With the current internalRequest type you can't set query parameters. I think we could add a QueryParams field, build the URL and add thoses query parameters in the Client.sendRequest method. That will avoid to have a special case in this method too (which is not pretty nice).

eskombro:
Great thanks for checking this! It is indeed unsafe, and your solution is clearly an improvement!

PR #44, opened 2020-06-12
Fix security issue with query parameters

alexisvisco:
Issue #43 describe the problem. This pull request resolve the security issue.


eskombro:
Thanks for this fix! Do you think it would be interesting to do the same for search request params?

PR #56, opened 2020-06-30
Make client List() parameters optional

### LINKED ISSUE #55: client.List() parameters need to be optional
Since the MeiliSearch HTTP server [changed from tide to atix](https://github.com/meilisearch/MeiliSearch/pull/601) they way it handles parameters that are empty has changed. As Go initializes every variable to a Default Zero value, when a parameter is not set to a value by the user, the Go SDK *will send the defult zero value to the server*. This leads to problems like #37 

For the client `List` method, the 3 parameters (`limit`, `offset` and `attributesToRetrieve`) are then initialized to their default zero values. If the user does not specify, for example some `attributesToRetrieve`, Go SDK is making a request with an empty list of `attibutesToRetrieve`. This parameters need to be added to the request only if the user specifies a value for them.

Test on master are failing because of this, as we can see for example [here](https://github.com/meilisearch/meilisearch-go/actions/runs/151697186)

GitHub actions arre nonetheless accepting those tests and validatign the check, which is a different issue and should be handled in #51

eskombro:
Issue detailed in #55 

`List` method parameters (`limit`, `offset` and `attributesToRetrieve`) should be optional, and shouldn't be sent on the request unless the user has set them to a specific value

Closes #55

### DIFF
--- a/client_documents.go
+++ b/client_documents.go
@@ -72,19 +72,26 @@ func (c clientDocuments) Deletes(identifier []string) (resp *AsyncUpdateID, err 
 }
 
 func (c clientDocuments) List(request ListDocumentsRequest, documentsPtr interface{}) error {
+
 	req := internalRequest{
-		endpoint:     "/indexes/" + c.indexUID + "/documents",
-		method:       http.MethodGet,
-		withRequest:  &request,
-		withResponse: documentsPtr,
-		withQueryParams: map[string]string{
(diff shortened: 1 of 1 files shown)


eskombro:
@alexisvisco I merged this PR as the tests where failing in master and this seemed a little urgent, but I requested your review to see if you agree with this! So feel free to comment, etc if you have a minute :)

PR #65, opened 2020-07-20
Remove sys-info route

### LINKED ISSUE #63: Remove sys-info routes
As stated in https://github.com/meilisearch/MeiliSearch/pull/810 and https://github.com/meilisearch/integration-guides/issues/36 the sys-info routes will be considered as deprecated in MeiliSearch v0.13.0. These routes should be removed from this SDK accordingly.

eskombro:
Removes sys-info route

Closes #63


curquiza:
If I'm not wrong this change is breakable! I add the label `breaking-change`, feel free to remove it if you don't agree.

PR #94, opened 2020-11-04
Update CI structure

curquiza:
The goal is to separate the different checks.
Why? Currenlty, on `master`, if the linter check fails, we don't know if the tests or the build would have failed. We have to fix the lint error and then wait for the next CI.
With this separation, the build, the linter, and the integration tests will run in different jobs so, independently in different runners.

What we will see now:

Capture d’écran 2020-11-04 à 19 22 25


curquiza:
bors try

.github/workflows/check.yml
```diff
@@ -0,0 +1,74 @@
...
+      - trying
+      - staging
+      - master
+
+jobs:
+  linter:
+    name: linter
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go 1.13
```
eskombro, line 17:
```suggestion
    - name: Set up Go 1.12
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,74 @@
...
+      run: /home/runner/go/bin/golint
+
+  integration_tests:
+    name: integration-tests
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go 1.13
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.13
```
eskombro, line 41:
```suggestion
        go-version: 1.12
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,74 @@
...
+      run: go test -v ./...
+
+  go_build:
+    name: go-build
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go 1.13
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.13
```
eskombro, line 63:
```suggestion
        go-version: 1.12
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,74 @@
...
+
+jobs:
+  linter:
+    name: linter
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go 1.13
+      uses: actions/setup-go@v1
+      with:
+        go-version: 1.13
```
eskombro, line 20:
```suggestion
        go-version: 1.12
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,74 @@
...
+        fi
+        go get -u golang.org/x/lint/golint
+    - name: Run linter
+      run: /home/runner/go/bin/golint
+
+  integration_tests:
+    name: integration-tests
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go 1.13
```
eskombro, line 38:
```suggestion
    - name: Set up Go 1.12
```

.github/workflows/check.yml
```diff
@@ -0,0 +1,74 @@
...
+    - name: MeiliSearch setup (latest version) with Docker
+      run: docker run -d -p 7700:7700 getmeili/meilisearch:latest ./meilisearch --master-key=masterKey --no-analytics=true
+    - name: Run integration tests
+      run: go test -v ./...
+
+  go_build:
+    name: go-build
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go 1.13
```
eskombro, line 60:
```suggestion
    - name: Set up Go 1.12
```

eskombro (changes requested):
This seems like a good idea, but it also triggers 3 times the environment set, docker image pull, etc.

I think this problem could be adressed by just altering the order of the checks: linter should be the last check, we don't want (need) to test linting before knowing if the code is even compilable. So the checks could go in this order:

1. build
2. Test
3. Lint

This would address a big part of the problem without having to run 3 CI at every push.

WDYT @curquiza ?

A not on this PR:

- The go.mod file defines the version compatibility. It is currently set to 1.12, so test shouldn't be done in a different environment (or we can upgrade the go.mod, but I think this should be consistent).

eskombro:
Suggestion:

We could add a separate CI on MERGE ONLY testing different OS like this one: https://github.com/mvdan/github-actions-golang#github-actions-for-go and different go versions

The keep a simple CI on PUSH just to check build + test + lint

eskombro (commented):
Suggestion:

We could add a separate CI on MERGE ONLY testing different OS like this one: https://github.com/mvdan/github-actions-golang#github-actions-for-go and different go versions

The keep a simple CI on PUSH just to check build + test + lint

curquiza:
The lint and the tests are 2 completely different and independent checks. Why should I wait for the tests are all green (a check that could be reallyyyy long for some SDKs) to see the errors of the linter? 
A test could fail but could have no link with the linter error. I should have the possibility to know all the errors at the same time. What is the advantage of hiding errors if we could avoid it with a more structured CI?

> This would address a big part of the problem without having to run 3 CI at every push.

It's not 3 CIs but 3 jobs 😉
What's the issue of running 3 jobs? Putting everything in the same job makes the job not really clear about its role, and less easy to read in the PR.
Here, as a contributor, I know clearly what is run: "linter", "integration-tests" and "go-build", and I know the separated result of each one.
Before, I just know "go-tests" was passed or not. Not really clear.

eskombro:
In this case, do you think we need a build job?

`go test ...` builds before testing, and if it can't perform a build, it throws a very clear error message:

Screenshot 2020-11-05 at 14 40 26

Build in go is almost instant.

curquiza:
And yes, the `integration-tests` should be done in different go versions, but in another PR maybe 🙂

curquiza:
> In this case, do you think we need a build job?

Indeed, the build job is not needed thanks to the tests, I can remove it! Do you agree with this removal?

eskombro:
> And yes, the `integration-tests` should be done in different go versions, but in another PR maybe 🙂

I'll create an issue about it if you prefer

curquiza:
bors try

eskombro (approved):
🌮

curquiza:
bors merge

PR #99, opened 2020-11-05
Run build and test on current go.mod version and latest stable go version

eskombro:
This PR introduces a 2 jobs test in the CI:
- one with the current go version of meilisearch-go module (1.12, set in `go.mod`)
- a second one with the latest stable go version (1.15 as of today).

It uses `include` in the GitHub Actions to add a tag system that uses the word `current` instead of `1.12` and `latest` instead of `1.15` to make it easy to automate GitHub and Bors checks by using always stable names. 

This way, for example, a job isn't called 

```
Check / integration-tests with go 1.12 version
```

but 

```
Check / integration-tests with go current version
```

For example, by changing the tests version from `1.12` -> `1.13` we won't need to update the repo nor the bors settings to do all the CI + checks, a simple version number change will work out-of-the-box! 🎉 

NOTE: It also removes some unused dependencies! 

Closes #99

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -15,9 +15,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Set up Go
-      uses: actions/setup-go@v1
+      uses: actions/setup-go@v2
     - name: Check out code into the Go module directory
-      uses: actions/checkout@v1
+      uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


eskombro:
bors try

eskombro:
bors try

curquiza (approved):
Awesome!!

PR #111, opened 2020-12-02
Update REAMDE according to the implicit index creation

curquiza:
Related to meilisearch/integration-guides#48

I cannot make more changes since this package needs a refacto.

With:

```
updateRes, err := client.Documents("books").AddOrUpdate(documents)
```

It now looks like the document is called `books` and no index is mentioned.

But not for this PR, this deserves another one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,15 +79,6 @@ func main() {
         APIKey: "masterKey",
     })
 
-    // Create an index if your index does not already exist
-    _, err := client.Indexes().Create(meilisearch.CreateIndexRequest{
-        UID: "books",
-    })
-    if err != nil {
-        fmt.Println(err)
-        os.Exit(1)
-    }
-
     documents := []map[string]interface{}{
         {"book_id": 123,  "title": "Pride and Prejudice"},
         {"book_id": 456,  "title": "Le Petit Prince"},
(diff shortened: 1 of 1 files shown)


eskombro (changes requested):
I do not like this result at all. There is no reference to the index, nor an example on how to use it. I find this readme in the actual state more confusing than helping!

If you want I can try to work on it a bit, or am open to any consideration. But wouldn't give this "approval" as it is :(

#NotInMyName :D

curquiza:
Ok, I'm really open to a suggestion to improve this Getting Starting without modifying the code (we will need a refacto, but not a priority for the moment)

Or we might change nothing at all for the moment, BUT keep in mind the \[Gettings Started of the docs\](https://docs.meilisearch.com/guides/introduction/quick_start_guide.html) will change soon and will remove the index creation step. So finding the right way to introduce this package without index manipulation will be an issue anyway.

So I'm waiting for your suggestion :+1:  :smile:

eskombro:
> So I'm waiting for your suggestion 👍  😄

Sounds like a challenge, so challenge accepted :)

curquiza:
Closing since it brings more confusion.
This package needs a refactor anyway to be more convenient to use.

PR #141, opened 2021-03-24
Fix health tests

curquiza:


curquiza:
@eskombro I don't know why but the `go get -v -t -d ./...` command updated the go.sum. Not sure this is what we want

eskombro (approved):
It just added the dependencies to the go.sum and since we are adding a new package in this PR [`github.com/valyala/fastjson`] it ads this package and its own dependencies to ours. It is totally ok from my POV

PR #150, opened 2021-05-27
Rewrite client

alallema:
**Description**
Start of SDK go modification:
- Rewriting of the Client that implements this method:
	- Index(uid string)
	- GetIndex(indexID string)
	- GetAllIndexes()
	- CreateIndex(config IndexConfig)
	- GetOrCreateIndex(config IndexConfig)
	- GetKeys()
	- GetAllStats()
	- CreateDump()
	- GetDumpStatus(dumpUID string)
	- Version()
	- GetVersion()
	- Health()
	- IsHealthy()
	- DeleteIndex(uid string)
- Rewriting of the Index that implements this method:
	- FetchInfo()
	- FetchPrimaryKey()
	- ChangePrimaryKey()
	- Delete()
	- AddDocuments()
	- UpdateDocuments()
	- GetDocument()
	- GetDocuments()
	- DeleteOneDocument()
	- DeletesDocuments()
	- DeleteAllDocuments()
	- GetAllUpdateStatus()
	- GetUpdateStatus()
	- Search()
	- GetSettings()
	- UpdateSettings()
	- ResetSettings()
	- GetStats()
	- WaitForPendingUpdate()
	- DefaultWaitForPendingUpdate()
- Removing of intermediates interfaces
- Rewriting of sample and Readme for the new usage
- Add tests for every method

**To do**
- Client
	- [x] Rewrite Client
	- [x] Add Test to Client
- Index 
	- [x] Rewrite Index
	- [x] Add Tests to Index
	- [x] Adding test for `WaitForPendingUpdate()` Issue [related](https://github.com/meilisearch/integration-guides/issues/1)
	- [x] Rewrite Document
	- [x] Add Tests to Document
	- [x] Rewrite Search
	- [x] Add Tests Search
	- [x] Rewrite Settings
	- [x] Add Tests Settings


client.go
```diff
@@ -1,328 +1,166 @@
...
-	return c.apiHealth
-}
-
-// Dumps return an APIDumps client.
-func (c *Client) Dumps() APIDumps {
-	return c.apiDumps
-}
+	Index(uid string) *Index
+	GetIndex(indexID string) (resp *Index, err error)
+	GetAllIndexes() (resp []Index, err error)
```
eskombro, line 24: Wouldn't it be more interesting if this method returned a slice of references to Index, so they are passed by reference and not by value?

```go
GetAllIndexes() (resp []*Index, err error)
```

client.go
```diff
@@ -1,328 +1,166 @@
...
+	CreateIndex(config IndexConfig) (resp *Index, err error)
+	GetOrCreateIndex(config IndexConfig) (resp *Index, err error)
```
eskombro, line 26: In `CreateIndex` and `GetOrCreateIndex`, the parameter `IndexConfig` could be also be passed as a reference instead of by value

index.go
```diff
@@ -0,0 +1,275 @@
...
+	UpdateDisplayedAttributes(request []string) (resp *AsyncUpdateID, err error)
+	ResetDisplayedAttributes() (resp *AsyncUpdateID, err error)
+	GetStopWords() (resp *[]string, err error)
+	UpdateStopWords(request []string) (resp *AsyncUpdateID, err error)
+	ResetStopWords() (resp *AsyncUpdateID, err error)
+	GetSynonyms() (resp *map[string][]string, err error)
+	UpdateSynonyms(request map[string][]string) (resp *AsyncUpdateID, err error)
+	ResetSynonyms() (resp *AsyncUpdateID, err error)
+	GetAttributesForFaceting() (resp *[]string, err error)
+	UpdateAttributesForFaceting(request []string) (resp *AsyncUpdateID, err error)
```
eskombro, line 66: Pass by reference? *[]string

index.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+var _ IndexInterface = &Index{}
+
+func newIndex(client *Client, uid string) *Index {
+	i := &Index{
+		UID:    uid,
+		client: client,
+	}
+	return i
```
eskombro, line 80: Unnecesary assignation
```suggestion
	return &Index{
		UID:    uid,
		client: client,
	}
```

index.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	if request.Matches {
+		searchPostRequestParams["matches"] = request.Matches
+	}
+	if len(request.FacetsDistribution) != 0 {
+		searchPostRequestParams["facetsDistribution"] = request.FacetsDistribution
+	}
+	if request.FacetFilters != nil {
+		searchPostRequestParams["facetFilters"] = request.FacetFilters
+	}
```
eskombro, line 261: All of these default values were hardcoded before. I can not come up with a good solution for this, but just as a question to think about, is there maybe any other way to store these default values somewhere instead of lost in the middle of the code?

For example: Limit => 20

index.go
```diff
@@ -0,0 +1,275 @@
...
+		time.Sleep(interval)
+	}
+}
+
+func (i Index) Search(request SearchRequest) (*SearchResponse, error) {
+	resp := &SearchResponse{}
+	searchPostRequestParams := map[string]interface{}{}
+	if request.Limit == 0 {
+		request.Limit = 20
+	}
```
eskombro, line 228: I don't understand this... If limit is 0, limit becomes 20?

index_documents.go
```diff
@@ -0,0 +1,162 @@
...
+		method:              http.MethodGet,
+		withRequest:         request,
+		withResponse:        response,
+		withQueryParams:     map[string]string{},
+		acceptedStatusCodes: []int{http.StatusOK},
+		functionName:        "GetDocuments",
+	}
+	if request.Limit != 0 {
+		req.withQueryParams["limit"] = strconv.FormatInt(request.Limit, 10)
+	}
```
eskombro, line 36: why the condition `request.Limit != 0`?

index_documents.go
```diff
@@ -0,0 +1,162 @@
...
+	if request.Offset != 0 {
```
eskombro, line 37: why the condition `request.Offset != 0`

eskombro (commented):
Amazing work @alallema !

I really think this is improving readability and usability at the same time. The general structure is simplified, without loosing any functionality and making it more friendly. I made a few comments on things that were not clear to me or that I think could be improved, LMKWYT :)

Good job 🎉

client.go
alallema: @eskombro,
I used to send the value instead of reference if the variable is expected to not be modified and it's not a big structure. Also I preferred this syntax:
``` go
	index, err := client.GetOrCreateIndex(meilisearch.IndexConfig{
		Uid: "indexUID",
	})
```
I can change it but it's used in the instantiation of the client too:
``` go
	var client = meilisearch.NewClient(meilisearch.ClientConfig{
		Host:      "http://127.0.0.1:7700",
		MasterKey: "masterKey",
	})
```
I can change and send references in every function if you think it's better.

index.go
alallema: I didn't modify this part of the code. But I think it is in case the limit is not specified in the `searchRequest` parameter the default value of meilisearch will be taken.

index_documents.go
alallema: Because in golang there are no optional parameters, all query parameters are passed via structures and checked if they exist. Maybe there is a better way to do it.

index.go
```diff
@@ -0,0 +1,275 @@
...
+	UpdateDistinctAttribute(request string) (resp *AsyncUpdateID, err error)
+	ResetDistinctAttribute() (resp *AsyncUpdateID, err error)
+	GetSearchableAttributes() (resp *[]string, err error)
+	UpdateSearchableAttributes(request []string) (resp *AsyncUpdateID, err error)
+	ResetSearchableAttributes() (resp *AsyncUpdateID, err error)
+	GetDisplayedAttributes() (resp *[]string, err error)
+	UpdateDisplayedAttributes(request []string) (resp *AsyncUpdateID, err error)
+	ResetDisplayedAttributes() (resp *AsyncUpdateID, err error)
+	GetStopWords() (resp *[]string, err error)
+	UpdateStopWords(request []string) (resp *AsyncUpdateID, err error)
```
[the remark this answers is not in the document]
alallema: Done

client.go
alallema: Done

client.go
alallema: Done

index.go
eskombro: Oh yes, you are right, this is because of the default zero values in Go. If you don't specify any limit, Go will initialize the struct and make it 0. So we modify it to 20. I remember we had a bug before and this is how it was fixed 👌🏽

eskombro: There will be no way to have a `Limit = 0`, but will someone use this kind of request? That is OK from my POV

index_documents.go
eskombro: Ok, that's all right, again de default zero values in Go that force this behavior. 👌🏽

eskombro (approved):
🎉

index.go
alallema: @eskombro 
We could state:
``` go
const (
    DefaultAttributesToRetrieve   []string = nil
    DefaultFacetFilters.          interface{} = nil
    DefaultLimit                  int64 = 0
    // with all default values
)
```
Then use:
``` go
	if request.Limit =! DefaultLimit {
		searchPostRequestParams["limit"] = request.Limit
	}
```

And it could be more explicit in:
```
	if request.Limit == DefaultLimit {
		request.Limit = 20
	}
```
too.
I'm not convinced of this, what did you think?

index.go
```diff
@@ -0,0 +1,275 @@
...
+	AddDocuments(documentsPtr interface{}) (resp *AsyncUpdateID, err error)
+	UpdateDocuments(documentsPtr interface{}) (*AsyncUpdateID, error)
+	AddDocumentsWithPrimaryKey(documentsPtr interface{}, primaryKey string) (resp *AsyncUpdateID, err error)
+	UpdateDocumentsWithPrimaryKey(documentsPtr interface{}, primaryKey string) (*AsyncUpdateID, error)
+	GetDocument(identifier string, documentPtr interface{}) error
+	GetDocuments(request DocumentsRequest, response interface{}) error
+	DeleteDocument(identifier string) (resp *AsyncUpdateID, err error)
+	DeleteDocuments(identifier []string) (resp *AsyncUpdateID, err error)
+	DeleteAllDocuments() (resp *AsyncUpdateID, err error)
+	Search(request SearchRequest) (*SearchResponse, error)
```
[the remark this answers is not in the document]
alallema: Done

index.go
eskombro: I agree, I am not a big fan, but at least is more readable for 2 reasons:

- It explains why theses values are there, by adding a name to this constants
- It regroups all of the hardcoded values at one place, probably at the beginning of the file.

This same principle could be applied by creating a hashmap, or a struct.

So even if I don't think this is an amazing solution, it still feels like an improvement :)

WDYT?

alallema: Same! I will do the change.
Thanks again @eskombro

README.md
```diff
@@ -67,43 +67,36 @@ NB: you can also download MeiliSearch from **Homebrew** or **APT**.
...
-    }
-
-    updateRes, err := client.Documents("books").AddOrUpdate(documents) // => { "updateId": 0 }
-    if err != nil {
-        fmt.Println(err)
-        os.Exit(1)
-    }
-
-    fmt.Println(updateRes.UpdateID)
+	client = meilisearch.NewClient(meilisearch.ClientConfig{
```
eskombro, line 77: Shouldn't `client` exist already (previously defined/initialized) or use := for this to work?

eskombro (changes requested):
Everything LGTM but from my POV there is a little mistake on the README, can you check?

🎉

eskombro (approved):
Amazing! 🎉

alallema:
bors merge

curquiza (approved):

alallema:
bors merge

PR #179, opened 2021-07-14
Fix codesample for Get Started Add documents section

sirodoht:
In this page: https://docs.meilisearch.com/learn/getting_started/quick_start.html#add-documents

Screenshot:
image

The highlighted line is not indented correctly (mixed spaces and tabs) while also `client` is not defined at this point, which is why I switched to `:=`

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -346,7 +346,7 @@ getting_started_add_documents_md: |-
   )
 
   func main() {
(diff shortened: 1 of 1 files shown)


alallema (approved):
Hi @sirodoht!
Thanks for this PR. You are absolutely right several errors have crept into the code_sample of the go. I fix some in this [PR](https://github.com/meilisearch/meilisearch-go/pull/176) but as it won't be released right away it's better to fix it now!

alallema:
bors merge

PR #184, opened 2021-07-22
Fix new search query in code sample

alallema:
Fix a lot of mistake in code sample

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -204,54 +204,54 @@ field_properties_guide_displayed_1: |-
   }
   client.Index("movies").UpdateDisplayedAttributes(&displayedAttributes)
 filtering_guide_1: |-
-  results, err := client.Index("movies").Search("Avengers", &meilisearch.SearchRequest{
+  resp, err := client.Index("movies").Search("Avengers", &meilisearch.SearchRequest{
     Filters: "release_date > \"795484800\"",
   })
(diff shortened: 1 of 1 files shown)


.code-samples.meilisearch.yaml
```diff
@@ -357,7 +351,7 @@ getting_started_add_documents_md: |-
     var movies []map[string]interface{}
     json.Unmarshal(byteValue, &movies)
 
-    update, err := client.Index("movies").AddDocuments(movies)
+    _, err := client.Index("movies").AddDocuments(movies)
```
bidoubiwa, line 354: In some samples you leave the variable even if it is not used in the same asmple 

For example: 

```
faceted_search_update_settings_1: |-
  response, err := client.Index("movies").UpdateAttributesForFaceting(&[]string{
    "director",
    "genres",
  })
```

How do you decide when you `_` the variable ?

alallema: You're right I think it better to leave the name everywhere.
Also, I realize that sometimes it's `result` and sometimes `response`.

alallema: Ah but this one it's the getting started. So I change it to be ready to use if you copy-paste the code

bidoubiwa: I think you're right!

bidoubiwa (approved):
LGTM

alallema:
bors merge

PR #190, opened 2021-08-18
Tests cleanup

gillesfabio:
👋 Meilisearch team,

Here's a small PR that cleans up tests.

* Use `t.Cleanup()` because calling `deleteAllIndexes()` isn't so DRY and safe
* Add missing error checks
* Remove dead code

This PR does NOT:

* Fix the `Health` test (it's still failing)
* Break compatibility
* Introduce any new dependencies
* Rename exiting variables

This PR doesn't really deserves any new GitHub issue. So I bypassed the contribution guide rules (yeah, bad I am).

Do whatever you want with these changes.

Keep making a great job! Meilisearch is amazing!


main_test.go
```diff
@@ -33,6 +34,17 @@ func deleteAllIndexes(client ClientInterface) (ok bool, err error) {
 	return true, nil
 }
 
+func cleanup(c ClientInterface) func() {
+	return func() {
+		_, _ = deleteAllIndexes(c)
+	}
+}
+
+func waitForPendingUpdate(t *testing.T, i *Index, u *AsyncUpdateID) {
```
alallema, line 43: There already is a function with a too similar name in the codebase `WaitForPendingUpdate` even if the other one is linked to the `Index` struct can you change this function name?

alallema (commented):
Thanks a lot, @gillesfabio, for this PR!
I had hesitated to use the fixtures of `testify` instead of the generated tests; maybe I had should.
Also, we should launch the MeiliSearch release really soon, so maybe I would merge your PR only after.
Thanks again for your contribution.

main_test.go
gillesfabio: I hesitated prefixing this function with `test` (so `testWaitForPendingUpdate`) to make a clear distinction between implementation and tests but that's a looong name for something often called and it's not so funky to read. It's like a namespacing patch. But, at least, it's clearer. Do you have any suggestions? I will follow you.

alallema: I totally agree with you! But as we are open source I prefer to be sure it's clear.
No suggestion it's great. Thanks again

gillesfabio: You're welcome! If you need any change before the merge, feel free to ping me. I'll fix/update/rebase.

About `testify` fixtures (I suppose you're talking about the test suite), to be honest, I'm not a huge fan of that package. I used it a lot in various projects and always felt uncomfortable with its API/test suite (it's a very subjective point of view). The test suite is more DRY when you need a lot of initialisation steps and data but a few helpers, like actually, it's cool too because more clear, more readable, less magic. Pros/cons, as usual.

Now, I rather prefer using `testing.T` + [google/go-cmp](https://github.com/google/go-cmp) or the wonderful [quicktest](https://github.com/frankban/quicktest) when I need to write/maintain more tests. [The `quicktest` suite](https://github.com/frankban/quicktest/blob/master/qtsuite/suite.go) is very simple and a pure joy to use.

Thanks for your fast replies.

alallema: Hi @gillesfabio!
Yes, I was talking about the test suite, thanks for the advice.
The release is taking longer than expected but it's out! I realize that I was not clear but I prefer if you could change the `waitForPendingUpdate` in `waitAndCheckUpdate`, `test_waitForPendingUpdate`, or another formulation even if it's a really long name but I prefer to have a reel distinction between both.
Thanks again and sorry for the delay

gillesfabio: Hi @alallema!

I'm sorry, my bad. Done! I renamed this function `testWaitForPendingUpdate`.

Hope it helps.

Have a nice day!

Issue #222, opened 2021-10-21
Rewrite `AddDocument` and `UpdateDocument` method with variadic parameter


alallema:
Currently there is two function `AddDocuments` and `AddDocumentsWithPrimaryKey`.
Both methods can be reunited in one with this prototype:
```
func (i Index) AddDocuments(documentsPtr interface{}, batchSize int, primaryKey ...string) (resp *AsyncUpdateID, err error)
```
Same for `UpdateDocuments` and `UpdateDocumentsWithPrimaryKey`. who can be rewrite with this prototype:
```
func (i Index) UpdateDocuments(documentsPtr interface{}, batchSize int, primaryKey ...string) (resp *AsyncUpdateID, err error)
```

alallema:
Hi @penthaapatel ,
Thanks for your interest, of course, you can work on this issue.
We prefer not assigning people to our issues because sometimes people ask to be assigned and never come back, which discourages the real volunteers contributors to open a PR to fix this issue.
We will accept and merge the first PR that fixes correctly and well implements the issue following our [contributing guidelines](https://github.com/meilisearch/meilisearch-java/blob/main/CONTRIBUTING.md).

PR #223, opened 2021-10-21
Rewrite `AddDocument` and `UpdateDocument` method with variadic parameter

penthaapatel:
Signed-off-by: Penthaa Patel <penthaapatel@gmail.com>

Fixes - #222

### DIFF
--- a/index_documents.go
+++ b/index_documents.go
@@ -51,14 +51,11 @@ func (i Index) GetDocuments(request *DocumentsRequest, resp interface{}) error {
 func (i Index) AddDocuments(documentsPtr interface{}, primaryKey ...string) (resp *AsyncUpdateID, err error) {
 	resp = &AsyncUpdateID{}
 	endpoint := ""
-	functionName := ""
 	if primaryKey == nil {
 		endpoint = "/indexes/" + i.UID + "/documents"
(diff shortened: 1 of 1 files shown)


penthaapatel:
@alallema Do we need to combine the functionality of 
 - all 3 functions `AddDocuments()`, `AddDocumentsWithPrimaryKey()` and `AddDocumentsInBatches()` into a single function? Or keep the `AddDocumentsInBatches()` as a separate function? 
 - similarly for all 3 functions `UpdateDocuments()`, `UpdateDocumentsWithPrimaryKey()` and `UpdateDocumentsInBatches()` into a single function? Or keep the `UpdateDocumentsInBatches()` as a separate function? 

Since combining all 3 functions into one will require changes in the current used test files and test cases too, so I just wanted to confirm before proceeding with this approach - if it would be okay? If yes, I will push further changes to this PR.

The PR that I have currently created combines the functionality of 
 - `AddDocuments()` and `AddDocumentsWithPrimaryKey()` into one function `AddDocuments()`, and 
 - `UpdateDocuments()` and `UpdateDocumentsWithPrimaryKey()` into one function `UpdateDocuments()`

PR #230, opened 2021-10-30
Make sure that `createdAt` and `updatedAt` are correctly update in `Index`

### LINKED ISSUE #209: Add `updatedAt` and `createdAt` to `Index`
When manipulating indexes (get/create/update), the MeiliSearch server returns the following information related to the index:
- `uid`: the uid of the index
- `primaryKey`: the primary key of the index
- `updateAt`: when the index has been updated
- `createdAt`: when the index has been created

This SDK, when manipulating indexes (get/create/update) returns an instance of `Index` which already contains the `primaryKey` and `uid` members.

It means we can currently get the `primaryKey` and the `uid`, but not `updatedAt` and `createdAt`.

TODO:
- [x] Add `updatedAt` and `createdAt` member in the `Index` class/struct (depending on the language).
- [ ] Update the variable in the code where `primaryKey` is already updated: Ex: when getting the index with `getIndex()`
- [ ] Add tests

⚠️ This issue is generated, it means `primaryKey`, `updatedAt`, and `createdAt` might be named differently in this package (ex: `primary_key` instead). Keep the already existing way of naming in this package to stay idiomatic with the language and the repository.
Also, if you are a maintainer, feel free to add any clarification and instruction about this issue related to the specific languages of the repo.

Related to https://github.com/meilisearch/integration-guides/issues/121

Sorry if this is already partially/completely implemented, feel free to let me know about the state of this issue in the repo.

theag3nt:
Added checks in related test cases to make sure that `createdAt` and `updatedAt` fields in `Index` are always filled when expected. These checks are currently passing, so the `Index` values returned should be correct already. 

Fixes #209

### DIFF
--- a/client_index_test.go
+++ b/client_index_test.go
@@ -133,6 +133,9 @@ func TestClient_CreateIndex(t *testing.T) {
 				if assert.NotNil(t, gotResp) {
 					require.Equal(t, tt.wantResp.UID, gotResp.UID)
 					require.Equal(t, tt.wantResp.PrimaryKey, gotResp.PrimaryKey)
+					// Make sure that timestamps are also retrieved
+					require.NotZero(t, gotResp.CreatedAt)
+					require.NotZero(t, gotResp.UpdatedAt)
 				}
 			}
 		})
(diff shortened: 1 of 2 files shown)


curquiza:
Hello @theag3nt 
Since the maintainer of the repo @alallema is on holiday, she will not able to review it before the end of the hacktoberfest. In case you are participating to, I add the `hacktoberfest-accepted` label for your work
Thanks again for contributing.

theag3nt:
The original issue mentions to also update these fields when `primaryKey` is updated. However it looks like the attempts to update `primaryKey` (and other `Index` fields in some cases) are ineffective because these methods use value receivers not pointers.

Let me know if these should be either:
- fixed, so the method result is the same as the instance the method was called on, e.g:
```go
i1 = c.Index("index")
i2 := i1.FetchInfo()
reflect.DeepEqual(i1, i2) // true
```
- removed, so they are not confusing

But it might make sense to do this in another merge request.

alallema (approved):
Hi @theag3nt,
Thanks for this PR and sorry for the delay.
Everything looks good to me.
And, you are right the method should be fixed and it could be done in another PR.

alallema:
bors merge

Issue #215, opened 2021-10-19
NDJSON/CSV methods to add and update documents


curquiza:
⚠️ This issue is generated, it means the nameing might be done differently in this package (ex: `add_documents_json` instead of `addDocumentsJson`). Keep the already existing way of naming in this package to stay idiomatic with the language and this repository.

📣 We strongly recommend doing multiple PRs to solve all the points of this issue

[MeiliSearch v0.23.0](https://github.com/meilisearch/MeiliSearch/releases/tag/v0.23.1) introduces two changes:
- new valid formats to push data files, additionally to the JSON format: CSV and NDJSON formats.
- it enforces the `Content-type` header for every route requiring a payload (`POST` and `PUT` routes)

---

Here are the expected changes to completely close the issue:

- [ ] Currently, the SDKs always send `Content-Type: application/json` to every request. Only the `POST` and `PUT` requests should send the `Content-Type: application/json` and not the `DELETE` and `GET` ones.

- [ ] Add the following methods and 🔥 **the associated tests** 🔥 to ADD the documents. Depending on the format type (`csv` or `ndjson`) the SDK should send `Content-Type: application/x-dnjson` or `Content-Type: text/csv`)
  - [ ] `addDocumentsJson(string docs, string primaryKey)`
  - [ ] `addDocumentsCsv(string docs, string primaryKey)`
  - [ ] `addDocumentsCsvInBatches(string docs, int batchSize, string primaryKey)`
  - [ ] `addDocumentsNdjson(string docs, string primaryKey)`
  - [ ] `addDocumentsNdjsonInBatches(string docs, int batchSize, string primaryKey)`

- [ ] Add the following methods and 🔥 **the associated tests** 🔥 to UPDATE the documents. Depending on the format type (`csv` or `ndjson`) the SDK should send `Content-Type: application/x-dnjson` or `Content-Type: text/csv`)
  - [ ] `updateDocumentsJson(string docs, string primaryKey)`
  - [ ] `updateDocumentsCsv(string docs, string primaryKey)`
  - [ ] `updateDocumentsCsvInBatches(string docs, int batchSize, string primaryKey)`
  - [ ] `updateDocumentsNdjson(string docs, string primaryKey)`
  - [ ] `updateDocumentsNdjsonInBatches(string docs, int batchSize, string primaryKey)`

`docs` are the documents sent as `String`
[...]

PR #235, opened 2021-11-01
Add methods for new document formats

theag3nt:
This completes the biggest part of #215 and adds the requested new methods for document insertion together with new test cases. Additionally some extra methods were added which receive `io.Reader` to provide a more idiomatic interface, especially for batch requests.

If this approach is approved, I can easily add the document update methods as well.

### DIFF
--- a/index_documents.go
+++ b/index_documents.go
@@ -1,6 +1,11 @@
 package meilisearch
 
 import (
+	"bufio"
+	"bytes"
+	"encoding/csv"
+	"fmt"
+	"io"
 	"math"
 	"net/http"
 	"reflect"
@@ -48,7 +53,7 @@ func (i Index) GetDocuments(request *DocumentsRequest, resp interface{}) error {
 	return nil
 }
 
-func (i Index) AddDocuments(documentsPtr interface{}, primaryKey ...string) (resp *AsyncUpdateID, err error) {
+func (i Index) addDocuments(documentsPtr interface{}, contentType string, primaryKey ...string) (resp *AsyncUpdateID, err error) {
 	resp = &AsyncUpdateID{}
 	endpoint := ""
 	if primaryKey == nil {
@@ -60,7 +65,7 @@ func (i Index) AddDocuments(documentsPtr interface{}, primaryKey ...string) (res
 	req := internalRequest{
 		endpoint:            endpoint,
 		method:              http.MethodPost,
-		contentType:         contentTypeJSON,
+		contentType:         contentType,
 		withRequest:         documentsPtr,
 		withResponse:        resp,
 		acceptedStatusCodes: []int{http.StatusAccepted},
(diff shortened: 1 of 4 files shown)


curquiza:
@theag3nt thanks for your PR. @alallema is on holiday and will review your PR once she's back

curquiza:
@theag3nt I see there are git conflicts, could you fix them before @alallema does the review 🙂

theag3nt:
Since I've implemented this a new Docker image has been released with breaking changes so currently the `main` branch is probably also failing tests.

I've tested everything again with the `getmeili/meilisearch:v0.23.1` Docker image and all tests are passing. I've also created a pull request with the fixes for v0.24.0 (#241).

alallema:
Like I said a PR was in progress for the release, I merge it so It should work properly now.

theag3nt:
Rebased on top of `main` once again.

index_documents.go
```diff
@@ -48,7 +55,7 @@ func (i Index) GetDocuments(request *DocumentsRequest, resp interface{}) error {
 	return nil
 }
 
-func (i Index) AddDocuments(documentsPtr interface{}, primaryKey ...string) (resp *AsyncUpdateID, err error) {
+func (i Index) addDocuments(documentsPtr interface{}, contentType string, primaryKey ...string) (resp *AsyncUpdateID, err error) {
```
alallema, line 58: It could be nice to rename this function `AddDocumentsRaw` and make it public?

alallema (commented):
Hey @theag3nt 

Thanks a lot for the contribution! I added a few comments and questions, but I really appreciate your time to make it possible to add these methods!
Thank you also for your patience, it took me a while to revise it because I had a lot of things to catch up on before. And last things I really enjoy all the comments in your code.

Thanks for contributing to MeiliSearch ❤️

index_documents.go
theag3nt: In that case I think we'll have to document which Content-Types are accepted (or limit the possible values with a custom type and predefined constants) and how the different document types are handled by `sendRequest`, otherwise there might be confusion about the method.

Do you have a specific use case in mind?

index_documents.go
```diff
@@ -106,6 +117,192 @@ func (i Index) AddDocumentsInBatches(documentsPtr interface{}, batchSize int, pr
 	return resp, nil
 }
 
+func (i Index) AddDocumentsCsv(documents string, primaryKey ...string) (resp *AsyncUpdateID, err error) {
+	// []byte avoids JSON conversion in Client.sendRequest()
+	return i.addDocuments([]byte(documents), contentTypeCSV, primaryKey...)
+}
```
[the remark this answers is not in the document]
theag3nt: I definitely agree that `[]byte` is the most natural type expected (apart from `io.Reader` but that's already covered), people usually won't hand over random types to a method which handles CSV/NDJSON files so an `interface{}` is probably not needed.

I'll change the method signature.

index_documents.go
alallema: Not really. You are right it will be easier to leave it private but if it is possible to rename it to avoid confusion?

alallema (approved):
Thanks a lot @theag3nt  for contributing to Meilisearch ❤️

alallema:
bors merge

PR #236, opened 2021-11-08
Add clean cache command in CONTRIBUTING.md

curquiza:
Before running the test, we should enforce this command. Indeed, because of the cache, running the same tests against two different versions of MeiliSearch will not behave as expected: the second run will give the exact same result as the second one.


alallema (approved):
Thanks, @curquiza !
LGTM

alallema:
bors merge

PR #243, opened 2021-11-18
Convert the dump status to an enum

### LINKED ISSUE #242: Convert the dump status field to an enum
**Description**
The Update type has an UpdateStatus enum, while the Dump type has a status as a string. Since this is an enum (according to the documentation, [link](https://docs.meilisearch.com/reference/api/dump.html#get-dump-status)), it's a good idea to convert this to any enum. By converting this field to an enum, the status fields have consistent behaviour.

roelofjan-elsinga:
This change helps to make the status field more consistent between "Update" and "Dump" and fixes #242.

The UpdateStatus field is an enum, but the DumpStatus field is a string. 
In the docs, these fields have a set number of possibilities and that's why it'd be proper to make DumpStatus an enum, rather than a string.

### DIFF
--- a/client_test.go
+++ b/client_test.go
@@ -230,13 +230,13 @@ func TestClient_GetDumpStatus(t *testing.T) {
 	tests := []struct {
 		name     string
 		client   *Client
-		wantResp []string
+		wantResp []DumpStatus
 		wantErr  bool
 	}{
 		{
 			name:     "TestGetDumpStatus",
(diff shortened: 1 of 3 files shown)


alallema (approved):
Thanks a lot, @roelofjan-elsinga for contributing to Meilisearch ❤️ 
LGTM! ✨

alallema:
bors merge

PR #262, opened 2022-01-27
Redesign update API to task API

alallema:
### Breaking Changes
All the actions on indexes are now asynchronous:
- `CreateIndex()`, `UpdateIndex()`, `DeleteIndex()` return a `Task` response instead of an `Index`.
- `index.Create` and `index.Delete` from index return a `task`.
- `WaitForPendingUpdate()`
    - is renamed into `WaitForTask`
    - the current `index.WaitForTask()` method call `/tasks/:uid`
- `index.GetUpdateStatus` is renamed `index.GetTask`
- `index.GetAllUpdateStatus` is renamed `index.GetTasks`

**Notes:** The only two methods that now return an `Index` are `client.Index()` and `client.GetIndex()`

### Enhancements
- new method `client.WaitForTask()` call `/tasks/:uid`
- new method `client.WaitForTask()`
- new method `client.GetTasks` that calls `/tasks`
- new method `client.GetTask` that calls `/tasks/:uid`

### DIFF
--- a/client.go
+++ b/client.go
@@ -23,6 +23,11 @@ type ClientConfig struct {
 	Timeout time.Duration
 }
 
+type waitParams struct {
+	Context  context.Context
+	Interval time.Duration
+}
+
 // ClientInterface is interface for all Meilisearch client
 type ClientInterface interface {
 	Index(uid string) *Index
@@ -42,8 +47,7 @@ type ClientInterface interface {
 	IsHealthy() bool
 	GetTask(taskID int64) (resp *Task, err error)
 	GetTasks() (resp *ResultTask, err error)
-	WaitForTask(ctx context.Context, interval time.Duration, task *Task) (*Task, error)
-	DefaultWaitForTask(task *Task) (*Task, error)
+	WaitForTask(task *Task, options ...waitParams) (*Task, error)
 }
 
 var _ ClientInterface = &Client{}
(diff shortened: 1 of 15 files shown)


client.go
```diff
@@ -38,6 +40,10 @@ type ClientInterface interface {
 	GetVersion() (resp *Version, err error)
 	Health() (*Health, error)
 	IsHealthy() bool
+	GetTask(taskID int64) (resp *Task, err error)
+	GetTasks() (resp *ResultTask, err error)
+	WaitForTask(ctx context.Context, interval time.Duration, taskID *Task) (*Task, error)
+	DefaultWaitForTask(taskID *Task) (*Task, error)
```
bidoubiwa, line 46: What is `defaultWaitForTask` ?

I see that it is when you provide the default values. Is this the naming convention in GO ?

client_request.go
```diff
@@ -134,7 +134,7 @@ func (c *Client) sendRequest(req *internalRequest, internalError *Error, respons
 		request.Header.Set("Content-Type", req.contentType)
 	}
 	if c.config.APIKey != "" {
-		request.Header.Set("Authorization", "Bearer " + c.config.APIKey)
+		request.Header.Set("Authorization", "Bearer "+c.config.APIKey)
```
bidoubiwa, line 137: The formater prefers no whitespaces around the `+` sign?

alallema: Me too, but the linter disagrees with us ...

client.go
alallema: It is the `WaitForTask` with default values. I don't think this naming is a specific convention in go. I can rename it if you have a better naming 🤩

bidoubiwa: in `waitForTask` you can add a structure the following way
```
type waitParams struct{
   interval: int,
   taskId: int
}
```

provide this as a parameter and in case of null use the default value?

```go
function waitForTask(taskId: int, params WaitParams) {
   var interval = 500 // default value
   if (params.interval != 0) {
     interval = params.interval
  }
  // etc ...
}

```

would something like that be possible?

client.go
```diff
@@ -38,6 +40,10 @@ type ClientInterface interface {
 	GetVersion() (resp *Version, err error)
 	Health() (*Health, error)
 	IsHealthy() bool
+	GetTask(taskID int64) (resp *Task, err error)
+	GetTasks() (resp *ResultTask, err error)
+	WaitForTask(ctx context.Context, interval time.Duration, taskID *Task) (*Task, error)
```
bidoubiwa, line 45:
```suggestion
	WaitForTask(ctx context.Context, interval time.Duration, task *Task) (*Task, error)
```

bidoubiwa (changes requested):

client.go
alallema: Yes, but you had to specify the last param at `nil` like:
```go
	client. waitForTask(0, nil)
```
There is no optional parameters in `golang` but I can use variadic args and do something like:
```go
func (c *Client) WaitForTask(task *Task, options ...waitParams) (*Task, error) {
	if (options == nil) {
		ctx, cancelFunc := context.WithTimeout(context.Background(), time.Second*5)
		defer cancelFunc()
		options[0].Context = ctx
		options[0].Interval = time.Millisecond*50
	}
... do waitForTask ...
}
```
And I can use it like:
```
	client. waitForTask(0)
	// OR
	client. waitForTask(0, &waitParams{ctx, time.Millisecond*10})
```
I'm not sure which one is the most elegant because in the second option I get the `waitParams` as an array.
What do you think?

bidoubiwa:
I like the variadic args! I'm not so used to goland so I hope it is something that is not choking to the user

bidoubiwa (approved):
LGTM 🔥

alallema:
> I like the variadic args! I'm not so used to goland so I hope it is something that is not choking to the user

Me too!

alallema:
bors merge

PR #264, opened 2022-02-01
Addition related to API keys

alallema:
### Breaking Changes

Granular management of API keys is now added to MeiliSearch. New methods have been created to manage this:
- `client.GetKey` get information about a specific API key.
- `client.CreateKey` create a new API key.
- `client.DeleteKey` delete an API key.
- `client.UpdateKey` update an API key.

Modification:
- `client.GetKeys`

### DIFF
--- a/types.go
+++ b/types.go
@@ -96,9 +96,18 @@ type ResultTask struct {
 // Keys allow the user to connect to the MeiliSearch instance
 //
 // Documentation: https://docs.meilisearch.com/learn/advanced/security.html#protecting-a-meilisearch-instance
-type Keys struct {
-	Public  string `json:"public,omitempty"`
-	Private string `json:"private,omitempty"`
+type Key struct {
+	Description string    `json:"description"`
+	Key         string    `json:"key,omitempty"`
+	Actions     []string  `json:"actions,omitempty"`
+	Indexes     []string  `json:"indexes,omitempty"`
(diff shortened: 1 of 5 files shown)


types.go
```diff
@@ -96,9 +96,29 @@ type ResultTask struct {
...
+
+// This structure is used to send the exact ISO-8601 time format managed by Meilisearch
+type KeyParsed struct {
+	Description string    `json:"description"`
+	Key         string    `json:"key,omitempty"`
+	Actions     []string  `json:"actions,omitempty"`
+	Indexes     []string  `json:"indexes,omitempty"`
+	CreatedAt   time.Time `json:"createdAt,omitempty"`
+	UpdatedAt   time.Time `json:"updatedAt,omitempty"`
+	ExpiresAt   *string   `json:"expiresAt"`
```
bidoubiwa, line 117: It is normal that only expiresAt has a string?

alallema: Yes, I will create an issue about it because there are two structs for handle key: `Key` and `KeyParsed`.
This one will not be used by the user it just allows me to recreate a `Key` with `expirationAt` that has the exact ISO-8601 and is a pointer because it could be set at `nil` in case you don't want to set and `expirationAt`

bidoubiwa (approved):
LGTM 🔥

alallema:
bors merge

PR #275, opened 2022-03-09
Adding GenerateTenantToken method to the client 

alallema:
## Tenant tokens

Introduction of the new method `GenerateTenantToken`  in order to facilitate the generation of the tenant token.

Related to:
- this issue: https://github.com/meilisearch/meilisearch/issues/1991
- this spec: https://github.com/meilisearch/specifications/pull/89

### DIFF
--- a/client.go
+++ b/client.go
@@ -54,7 +54,7 @@ type ClientInterface interface {
 	GetTask(taskID int64) (resp *Task, err error)
 	GetTasks() (resp *ResultTask, err error)
 	WaitForTask(task *Task, options ...waitParams) (*Task, error)
-	GenerateTenantToken(searchRules interface{}, options *TenantTokenOptions) (resp string, err error)
+	GenerateTenantToken(searchRules map[string]interface{}, options *TenantTokenOptions) (resp string, err error)
 }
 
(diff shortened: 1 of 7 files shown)


client.go
```diff
@@ -327,18 +327,18 @@ func (c *Client) WaitForTask(task *Task, options ...waitParams) (*Task, error) {
 // ApiKey options is the API key parent of the token. If you leave it empty the client API Key will be used.
 func (c *Client) GenerateTenantToken(SearchRules interface{}, Options *TenantTokenOptions) (resp string, err error) {
 	// Validate the arguments
-	if (SearchRules == nil) {
+	if SearchRules == nil {
```
bidoubiwa, line 330: Is it not possible in this case to check for these types? Because if `searchRules` is a integer it will not enter the if condition.
Same for the other if conditions

alallema: I change the type of `searchRules`. What do you mean by the other if conditions?

bidoubiwa: Is it not useful to check that `searchRules` is not an integer for example? Why do we only check if it is a nil ?

bidoubiwa (approved):
LGTM 🔥

PR #279, opened 2022-04-02
Feature/Analytics

brunoocasali:
- Add `User-Agent` inside the pre-defined headers

Add Go support as requested here meilisearch/integration-guides#150

I tried other options, but they do not fit so well:

- https://pkg.go.dev/debug/buildinfo (does not work very well :/ has a lot of issues https://github.com/golang/go/issues/29814, https://github.com/golang/go/issues/33975
- https://github.com/ahmetb/govvv (I think it is not an option since it seems to be not maintained anymore).

### DIFF
--- a/version.go
+++ b/version.go
@@ -1,0 +1,13 @@
+package meilisearch
+
+import "fmt"
+
+const VERSION = "0.19.1"
+
+func GetQualifiedVersion() (qualifiedVersion string) {
+	return getQualifiedVersion(VERSION)
+}
+
+func getQualifiedVersion(version string) (qualifiedVersion string) {
+	return fmt.Sprintf("Meilisearch Go (v%s)", version)
+}
--- a/client_request.go
+++ b/client_request.go
@@ -137,6 +137,8 @@ func (c *Client) sendRequest(req *internalRequest, internalError *Error, respons
 		request.Header.Set("Authorization", "Bearer "+c.config.APIKey)
 	}
 
+	request.Header.Set("User-Agent", GetQualifiedVersion())
+
 	// request is sent
 	if c.config.Timeout != 0 {
 		err = c.httpClient.DoTimeout(request, response, c.config.Timeout)
--- a/version_test.go
+++ b/version_test.go
@@ -1,0 +1,30 @@
+package meilisearch
+
+import (
+	"testing"
+	"fmt"
+	"regexp"
+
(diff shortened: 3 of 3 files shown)


brunoocasali:
@alallema, I had some trouble testing the presence of the `User-Agent` header. 
Feel free to merge like this (I can create an issue about it in the future) or point me out how to do that without changing many files 😆. 

One thing I thought was to add the header into the `Config`, but then the `Config` will need to be modified in the `NewClient` func, to make sure the header will be there.

Let me know your thoughts :)

version.go
```diff
@@ -0,0 +1,13 @@
...
+
+const VERSION = "0.19.1"
+
+func GetQualifiedVersion() (qualifiedVersion string) {
+	return getQualifiedVersion(VERSION)
+}
+
+func getQualifiedVersion(version string) (qualifiedVersion string) {
+	return fmt.Sprintf("Meilisearch Go (v%s)", version)
+}
```
alallema, line 13: What is the point of writing to function instead of one? Is it for having a public and a private one?

alallema (approved):
Except for one question LGTM! 🌮

alallema:
> @alallema, I had some trouble testing the presence of the User-Agent header.
Feel free to merge like this (I can create an issue about it in the future) or point me out how to do that without changing many files 😆.
One thing I thought was to add the header into the Config, but then the Config will need to be modified in the NewClient func, to make sure the header will be there - I had some trouble doing that too.
Let me know your thoughts :)

I think the best thing to do is to leave it as it is for now until a better solution is found if it exists or to open a way to ask for the community's help in this matter

version.go
brunoocasali: Yeah, I could add just one function but I was thinking about how to test it, so I created two functions one with the possibility to add a version (private) and another used in the app (public).

This way I could test the format and I could guarantee that the user will always send the right value "VERSION" to the function.
Because I could let this responsibility to the caller:

```go
	request.Header.Set("User-Agent", GetQualifiedVersion(VERSION))
```

But how can I ensure by tests the user is using the `VERSION` and not something else?

PS: I still need to understand more about the Go concepts to improve this 😅, feel free to point a better solution :)

alallema:
bors merge

alallema (approved):

alallema:
bors merge

PR #281, opened 2022-04-05
Fix expired at in UpdateKey method

alallema:
Following this discussion https://github.com/meilisearch/product/discussions/403 
Need to add some Test


alallema:
Closed fix by #313

PR #290, opened 2022-04-27
Ensure nested field support

alallema:
Add tests on nested fields to ensure it is still maintained in Meilisearch.

### DIFF
--- a/index_search_test.go
+++ b/index_search_test.go
@@ -919,13 +919,13 @@ func TestIndex_SearchWithSort(t *testing.T) {
 
 func TestIndex_SearchOnNestedFileds(t *testing.T) {
 	type args struct {
-		UID        string
-		PrimaryKey string
-		client     *Client
-		query      string
-		request    SearchRequest
+		UID                 string
+		PrimaryKey          string
+		client              *Client
+		query               string
+		request             SearchRequest
 		searchableAttribute []string
-		sortableAttribute []string
+		sortableAttribute   []string
 	}
 	tests := []struct {
 		name string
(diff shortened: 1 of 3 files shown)


index_search_test.go
```diff
@@ -916,3 +916,172 @@ func TestIndex_SearchWithSort(t *testing.T) {
...
+					},
+				},
+				NbHits:           1,
+				Offset:           0,
+				Limit:            20,
+				ExhaustiveNbHits: false,
+			},
+		},
+		{
+			name: "TestIndexBasicSearchOnNestedFieldsWithCustomClient",
```
bidoubiwa, line 956: what is a custom client?

index_search_test.go
```diff
@@ -916,3 +916,172 @@ func TestIndex_SearchWithSort(t *testing.T) {
...
+				searchableAttribute: []string{
+					"title", "info.comment",
+				},
+			},
+			want: &SearchResponse{
+				Hits: []interface{}{
+					map[string]interface{}{
+						"id": float64(5), "title": "The Hobbit",
+					},
+				},
```
bidoubiwa, line 1014: should you test the nested object as well?

main_test.go
```diff
@@ -133,6 +133,33 @@ func SetUpBasicIndex(indexUID string) {
 	}
 }
 
+func SetUpIndexWithNestedFields(indexUID string) {
+	client := NewClient(ClientConfig{
+		Host:   "http://localhost:7700",
+		APIKey: masterKey,
```
bidoubiwa, line 139:
```suggestion
		Host: "http://localhost:7700",
		APIKey: masterKey,
```

bidoubiwa (changes requested):

index_search_test.go
alallema: We currently use the `fasthttp` library to create an `http` client but in go you have the native `net/http` library to create a client. We offer the possibility for the user to generate his own client with `net/http` or another library.

index_search_test.go
alallema: Done !

PR #291, opened 2022-04-27
Add new search parameters highlightPreTag, highlightPostTag and cropMarker

alallema:
As per https://github.com/meilisearch/meilisearch/issues/2214 new search parameters are introduced: 

- `highlightPreTag`
- `highlightPostTag`
- `cropMarker`

### DIFF
--- a/index_search.go
+++ b/index_search.go
@@ -38,6 +38,15 @@ func (i Index) Search(query string, request *SearchRequest) (*SearchResponse, er
 	if request.CropLength != 0 {
 		searchPostRequestParams["cropLength"] = request.CropLength
 	}
+	if request.CropMarker != "" {
+		searchPostRequestParams["cropMarker"] = request.CropMarker
+	}
+	if request.HighlightPreTag != "" {
+		searchPostRequestParams["highlightPreTag"] = request.HighlightPreTag
+	}
+	if request.HighlightPostTag != "" {
+		searchPostRequestParams["highlightPostTag"] = request.HighlightPostTag
+	}
 	if len(request.AttributesToRetrieve) != 0 {
 		searchPostRequestParams["attributesToRetrieve"] = request.AttributesToRetrieve
 	}
--- a/index_search_test.go
+++ b/index_search_test.go
@@ -100,7 +100,7 @@ func TestIndex_Search(t *testing.T) {
 			want: &SearchResponse{
 				Hits: []interface{}{
 					map[string]interface{}{
-						"book_id": float64(1), "title": "Alice In Wonderland",
+						"book_id": float64(123), "title": "Pride and Prejudice",
 					},
 				},
 				NbHits:           20,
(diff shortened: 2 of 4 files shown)


index_search_test.go
```diff
@@ -164,13 +164,16 @@ func TestIndex_Search(t *testing.T) {
 				query:  "to",
 				request: SearchRequest{
 					AttributesToCrop: []string{"title"},
-					CropLength:       7,
+					CropLength:       2,
```
bidoubiwa, line 167: is this indentation normal? It feels like it is one space different with [

index_search_test.go
```diff
@@ -179,6 +182,87 @@ func TestIndex_Search(t *testing.T) {
...
+					},
+				},
+				NbHits:           2,
+				Offset:           0,
+				Limit:            1,
+				ExhaustiveNbHits: false,
+			},
+		},
+		{
+			name: "TestIndexSearchWithAttributeToHighlightWithCustomPreAndPostTag",
```
bidoubiwa, line 239:
```suggestion
			name: "TestIndexSearchWithCustomPreAndPostHighlightTags",
```

index_search.go
```diff
@@ -38,6 +38,15 @@ func (i Index) Search(query string, request *SearchRequest) (*SearchResponse, er
 	if request.CropLength != 0 {
 		searchPostRequestParams["cropLength"] = request.CropLength
 	}
+	if request.CropMarker != "" {
+		searchPostRequestParams["cropMarker"] = request.CropMarker
+	}
+	if request.HighlightPreTag != "" {
```
brunoocasali, line 44: There is a way to automate this kind of thing, like serialize somehow?

index_search_test.go
```diff
@@ -100,7 +100,7 @@ func TestIndex_Search(t *testing.T) {
 			want: &SearchResponse{
 				Hits: []interface{}{
 					map[string]interface{}{
-						"book_id": float64(1), "title": "Alice In Wonderland",
+						"book_id": float64(123), "title": "Pride and Prejudice",
```
brunoocasali, line 103: Can you give some idea about this `float64(...)`? Why do you need to convert it to float?

index_search.go
alallema: I would love it! I didn't find any other ways

index_search_test.go
alallema: This is the default behavior of the deserializer in golang you can read in the [documentation](https://pkg.go.dev/encoding/json#Unmarshal)

> To unmarshal JSON into an interface value, Unmarshal stores one of these in the interface value:
bool, for JSON booleans
float64, for JSON numbers
string, for JSON strings
[]interface{}, for JSON arrays
map[string]interface{}, for JSON objects
nil for JSON null

And by default, every number declare are `int` so in:
```
	map[string]interface{}{
		"book_id": 1,
	}
```
`book_id` will be an int that is why I convert it. I didn't find a better way to do it.

brunoocasali (approved):
Awesome! (go tests are a little bit harder to read for me yet, but I think they are ok) 👯‍♂️

alallema:
bors merge

PR #294, opened 2022-04-28
Add typo tolerance settings

### LINKED ISSUE #288: Add support to the typo tolerance customization
⚠️ This issue is generated, it means the examples and the namings do not necessarily correspond to the language of this repository.
Also, if you are a maintainer, feel free to add any clarification and instruction about this issue.

Related to:
- parent issue: https://github.com/meilisearch/integration-guides/issues/190
- specification: https://github.com/meilisearch/specifications/pull/117

----

Add a `typoTolerance` index settings resource to manage customization of the typo tolerance feature at the index level.

- Expose `typoTolerance` object on `/indexes/:index_uid/settings` endpoints.
- Add `GET/POST/DELETE` - `/indexes/:index_uid/settings/typo-tolerance` endpoints.

The `typoTolerance` object definition is made of the following properties:
  - `enabled` type (`boolean`)
  - `disableOnAttributes` type (`array[string]`)
  - `disableOnWords` type (`array[string]`)
  - `minWordSizeForTypos` type (`object`)
    - `oneTypo` type (`int`)
    - `twoTypos` type (`int`)

More information about what each field does could be [checked in the spec.](https://github.com/meilisearch/specifications/pull/117)

Check the other customization methods for reference like: `synonyms`, `searchableAttributes`, `filterableAttributes`.

TODO:
- [x] Add the new methods to customize `typo-tolerance`
  - [x] GET `/indexes/:index_uid/settings/typo-tolerance`
  - [x] POST `/indexes/:index_uid/settings/typo-tolerance`
  - [x] DELETE `/indexes/:index_uid/settings/typo-tolerance`
- [x] Add tests

alallema:
This PR introduces the new setting: [typoTolerance](https://github.com/meilisearch/specifications/pull/117)

new methods: 
`Index.GetTypoTolerance()`
`Index.UpdateTypoTolerance(params)`
`Index.ResetTypoTolerance()`

closed #288

### DIFF
--- a/index_settings.go
+++ b/index_settings.go
@@ -445,3 +445,52 @@ func (i Index) ResetSortableAttributes() (resp *Task, err error) {
 	}
 	return resp, nil
 }
+
+func (i Index) GetTypoTolerance() (resp *TypoTolerance, err error) {
+	resp = &TypoTolerance{}
+	req := internalRequest{
+		endpoint:            "/indexes/" + i.UID + "/settings/typo-tolerance",
+		method:              http.MethodGet,
+		withRequest:         nil,
+		withResponse:        resp,
+		acceptedStatusCodes: []int{http.StatusOK},
+		functionName:        "GetTypoTolerance",
+	}
+	if err := i.client.executeRequest(req); err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+func (i Index) UpdateTypoTolerance(request *TypoTolerance) (resp *Task, err error) {
+	resp = &Task{}
+	req := internalRequest{
+		endpoint:            "/indexes/" + i.UID + "/settings/typo-tolerance",
+		method:              http.MethodPost,
+		contentType:         contentTypeJSON,
+		withRequest:         &request,
+		withResponse:        resp,
+		acceptedStatusCodes: []int{http.StatusAccepted},
+		functionName:        "UpdateTypoTolerance",
+	}
+	if err := i.client.executeRequest(req); err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+func (i Index) ResetTypoTolerance() (resp *Task, err error) {
+	resp = &Task{}
+	req := internalRequest{
+		endpoint:            "/indexes/" + i.UID + "/settings/typo-tolerance",
+		method:              http.MethodDelete,
+		withRequest:         nil,
+		withResponse:        resp,
+		acceptedStatusCodes: []int{http.StatusAccepted},
+		functionName:        "ResetTypoTolerance",
+	}
+	if err := i.client.executeRequest(req); err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
--- a/index_settings_test.go
+++ b/index_settings_test.go
@@ -1322,11 +1322,11 @@ func TestIndex_UpdateSettings(t *testing.T) {
 					},
 					TypoTolerance: &TypoTolerance{
 						Enabled: true,
-						MinWordSizeForTypos: MinWordSizeForTypo{
+						MinWordSizeForTypos: MinWordSizeForTypos{
 							OneTypo:  7,
 							TwoTypos: 10,
 						},
-						DisableOnWords: []string{},
+						DisableOnWords:      []string{},
 						DisableOnAttributes: []string{},
 					},
 				},
@@ -1376,11 +1376,11 @@ func TestIndex_UpdateSettings(t *testing.T) {
 					},
 					TypoTolerance: &TypoTolerance{
 						Enabled: true,
-						MinWordSizeForTypos: MinWordSizeForTypo{
+						MinWordSizeForTypos: MinWordSizeForTypos{
 							OneTypo:  7,
 							TwoTypos: 10,
 						},
-						DisableOnWords: []string{},
+						DisableOnWords:      []string{},
 						DisableOnAttributes: []string{},
 					},
 				},
@@ -1898,7 +1898,7 @@ func TestIndex_UpdateSettingsOneByOne(t *testing.T) {
 					},
 					TypoTolerance: &TypoTolerance{
 						Enabled: true,
-						MinWordSizeForTypos: MinWordSizeForTypo{
+						MinWordSizeForTypos: MinWordSizeForTypos{
 							OneTypo:  7,
 							TwoTypos: 10,
 						},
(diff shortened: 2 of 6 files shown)


index_settings.go
```diff
@@ -445,3 +445,52 @@ func (i Index) ResetSortableAttributes() (resp *Task, err error) {
...
+	return resp, nil
+}
+
+func (i Index) UpdateTypoTolerance(request *TypoTolerance) (resp *Task, err error) {
+	resp = &Task{}
+	req := internalRequest{
+		endpoint:            "/indexes/" + i.UID + "/settings/typo-tolerance",
+		method:              http.MethodPost,
+		contentType:         contentTypeJSON,
+		withRequest:         &request,
```
brunoocasali, line 471: Can you enlighten me on this one?

index_settings_test.go
```diff
@@ -378,6 +380,47 @@ func TestIndex_GetSortableAttributes(t *testing.T) {
 	}
 }
 
+func TestIndex_GetTypoTolerance(t *testing.T) {
+	type args struct {
+		UID    string
+		client *Client
+	}
+	tests := []struct {
```
brunoocasali, line 388: This particular struct is different in every test? I was thinking of a way to make it reusable.

index_settings_test.go
```diff
@@ -378,6 +380,47 @@ func TestIndex_GetSortableAttributes(t *testing.T) {
...
+		name     string
+		args     args
+		wantResp *TypoTolerance
```
brunoocasali, line 391: What about just `response` or `expectation`?

index_settings_test.go
```diff
@@ -378,6 +380,47 @@ func TestIndex_GetSortableAttributes(t *testing.T) {
...
+				UID:    "indexUID",
+				client: customClient,
+			},
+			wantResp: &defaultTypoTolerance,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			SetUpIndexForFaceting()
+			c := tt.args.client
```
brunoocasali, line 413: It is a common practice to define variables like this in Go?

index_settings_test.go
```diff
@@ -858,6 +903,59 @@ func TestIndex_ResetSortableAttributes(t *testing.T) {
...
+			wantTask: &Task{
+				UID: 1,
+			},
+			wantResp: &defaultTypoTolerance,
+		},
+		{
+			name: "TestIndexResetTypoToleranceWithCustomClient",
+			args: args{
+				UID:    "indexUID",
+				client: customClient,
```
brunoocasali, line 932: Do you see a good value in keeping two same tests just changing the client? Since they implement the same `get`, `post`, `put`, `patch`, and `delete` HTTP methods, we can trust they will be ok in every test.

Think about this, and maybe we could remove the `customClient` testing from most of the tests. This will simplify a lot of the test suite without losing trustability.

index_settings_test.go
```diff
@@ -1222,6 +1320,15 @@ func TestIndex_UpdateSettings(t *testing.T) {
 					SortableAttributes: []string{
 						"title",
 					},
+					TypoTolerance: &TypoTolerance{
```
brunoocasali, line 1323: Do you need to check in every assertion if the object is equal to another object? What do you think about making the assertions with the part you need?

I saw a lot of changed tests with `TypoTolerance` stuff which did not concern them!

types.go
```diff
@@ -36,6 +36,21 @@ type Settings struct {
...
+// TypoTolerance is the type that represents the typo tolerance setting in Meilisearch
+type TypoTolerance struct {
+	Enabled             bool               `json:"enabled,omitempty"`
+	MinWordSizeForTypos MinWordSizeForTypo `json:"minWordSizeForTypos,omitempty"`
+	DisableOnWords      []string           `json:"disableOnWords,omitempty"`
+	DisableOnAttributes []string           `json:"disableOnAttributes,omitempty"`
+}
+
+// MinWordSizeForTypo is the type that represents the minWordSizeForTypo setting in the typo tolerance setting in Meilisearch
+type MinWordSizeForTypo struct {
```
brunoocasali, line 51: Is intentional to have the name `MinWordSizeForTypo` instead of `MinWordSizeForTypos` for your internal struct type?

types_easyjson.go
```diff
@@ -163,7 +163,170 @@ func (v *UpdateIndexRequest) UnmarshalJSON(data []byte) error {
 func (v *UpdateIndexRequest) UnmarshalEasyJSON(l *jlexer.Lexer) {
 	easyjson6601e8cdDecodeGithubComMeilisearchMeilisearchGo1(l, v)
 }
-func easyjson6601e8cdDecodeGithubComMeilisearchMeilisearchGo2(in *jlexer.Lexer, out *TenantTokenOptions) {
+func easyjson6601e8cdDecodeGithubComMeilisearchMeilisearchGo2(in *jlexer.Lexer, out *TypoTolerance) {
```
brunoocasali, line 166: Can you give some context behind this change?

brunoocasali (commented):
I have some comments, since this is one of my "first" GoLang reviews :)

types.go
alallema: No! It's a typo ... 😬

index_settings_test.go
alallema: The tests were autogenerated with `GoLand` but this soft is not free and I just use it for this purpose. I think It could be nice to find a free solution and add it to the `CONTRIBUTING`.

index_settings_test.go
alallema: No, I think you are right, keep a test for the `customClient` in the client test should be enough. I will open an issue about it.

index_settings_test.go
alallema: Same, as it is autogenerated I didn't change the name of the variable

index_settings_test.go
alallema: Not so much, but same it's because it's autogenerated

index_settings_test.go
alallema: I don't understand what you mean. Can you explain a little more?

index_settings.go
alallema: `Object` doesn't really exist in go, so to send multiple arguments it's easier to use struct. For this purpose, we create this struct with all data we need to process a request:
```go
type internalRequest struct {
	endpoint    string
	method      string
	contentType string

	withRequest     interface{}
	withResponse    interface{}
	withQueryParams map[string]string

	acceptedStatusCodes []int

	functionName string
}
```
We send the struct to the eexecuteRequest` function with all the data needed, and a `interface{}` in `withRequest` who will be fill with the response from the API.

index_settings_test.go
brunoocasali: Yes, but I think the answer will be "the tests are autogenerated" haha.

What I mean is, if in the test "update typo tolerance using settings" we just update one field, we don't need to assert every other field, just we want to be changed.

brunoocasali (approved):

PR #310, opened 2022-06-30
Apply tasks changes

alallema:
# Pull Request

Related to: https://github.com/meilisearch/integration-guides/issues/205

## What does this PR do?

Breaking because enforces the users to use Meilisearch v0.28.0

### Changes

- Remove `GET /indexes/:indexUid/tasks`. Use `GET /tasks?indexUid=:indexUid` instead.
- Remove `GET /indexes/:indexUid/tasks/:taskUid`. Use `GET /tasks/:taskUid` instead.
- Possibility to filter the results:
  - In the endpoint `GET /tasks` we can filter by: `type`, `status` and `indexUid`.
  - We need to join the array values when filter by `status` like this: `enqueued,processing`.
- [x] `Index.GetTask` now use `GET /tasks/:task_uid`
- [x] `Index.GetTasks` now use `GET/tasks?indexUid=movies`
- [x] Rename task `uid` field to `taskUid`.
- [x] `Index.GetTasks` accept pagination metadata, added limit (default: 20), from.
- [x] `Client.GetTasks` accept pagination metadata, added limit (default: 20), from.


index.go
```diff
@@ -158,32 +158,39 @@ func (i Index) GetStats() (resp *StatsIndex, err error) {
...
+	return i.client.GetTask(taskUID)
 }
 
-func (i Index) GetTasks() (resp *ResultTask, err error) {
-	resp = &ResultTask{}
+func (i Index) GetTasks(param *TasksQuery) (resp *TaskResult, err error) {
+	resp = &TaskResult{}
 	req := internalRequest{
-		endpoint:            "/indexes/" + i.UID + "/tasks",
+		endpoint:            "/tasks?indexUid=" + i.UID,
```
brunoocasali, line 168: I think you don't need this `?indexUid=" + i.UID` since you will get the value from `withQueryParams["indexUid"]` correct?

index.go
```diff
@@ -158,32 +158,39 @@ func (i Index) GetStats() (resp *StatsIndex, err error) {
...
+	}
+	if param != nil && param.Status != "" {
+		req.withQueryParams["status"] = param.Status
+	}
+	if param != nil && param.Type != "" {
+		req.withQueryParams["type"] = param.Type
+	}
+	if param != nil && len(param.IndexUID) != 0 {
+		param.IndexUID = append(param.IndexUID, i.UID)
+		req.withQueryParams["indexUid"] = strings.Join(param.IndexUID, ",")
```
brunoocasali, line 190: The import `strings` is missing :)

index.go
```diff
@@ -158,32 +158,39 @@ func (i Index) GetStats() (resp *StatsIndex, err error) {
...
-		endpoint:            "/indexes/" + i.UID + "/tasks",
+		endpoint:            "/tasks?indexUid=" + i.UID,
 		method:              http.MethodGet,
 		withRequest:         nil,
 		withResponse:        &resp,
+		withQueryParams:     map[string]string{},
 		acceptedStatusCodes: []int{http.StatusOK},
 		functionName:        "GetTasks",
 	}
+	if param != nil && param.Limit != 0 {
```
brunoocasali, line 176: Is there any way to re-use the code from `client.go` and `index.go`?

client.go
```diff
@@ -273,16 +274,32 @@ func (c *Client) GetTask(taskID int64) (resp *Task, err error) {
...
+		req.withQueryParams["limit"] = strconv.FormatInt(param.Limit, 10)
+	}
+	if param != nil && param.From != 0 {
+		req.withQueryParams["from"] = strconv.FormatInt(param.From, 10)
+	}
+	if param != nil && param.Status != "" {
+		req.withQueryParams["status"] = param.Status
+	}
+	if param != nil && param.Type != "" {
+		req.withQueryParams["type"] = param.Type
```
brunoocasali, line 298: I think you can define the map inline, and after that, remove the nulls or empties like this:

```go
map[string]string{
   "type": len(param.IndexUID) != 0 ? strings.Join(param.IndexUID, ",") : nil, 
   "limit": ...
}
```

brunoocasali:
@alallema I noticed there is some linter issues to be fixed as well, I'm not sure if you'll do that in this PR or not :)

.code-samples.meilisearch.yaml
```diff
@@ -62,11 +62,11 @@ search_post_1: |-
 get_task_1: |-
   client.GetTask(1);
 get_all_tasks_1: |-
-  client.GetTasks();
+  client.GetTasks(nil);
```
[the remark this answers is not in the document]
alallema: We can use variadic parameters but I don't think it's a good way to handle it, even if it works it's made for it.

client.go
alallema: You cannot put logic inline in a struct declaration in go or I don't know how

client.go
```diff
@@ -273,16 +274,32 @@ func (c *Client) GetTask(taskID int64) (resp *Task, err error) {
...
 	req := internalRequest{
 		endpoint:            "/tasks",
 		method:              http.MethodGet,
 		withRequest:         nil,
 		withResponse:        &resp,
+		withQueryParams:     map[string]string{},
 		acceptedStatusCodes: []int{http.StatusOK},
 		functionName:        "GetTasks",
 	}
+	if param != nil && param.Limit != 0 {
```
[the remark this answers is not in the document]
alallema: I agree but for that, I should declare `limit` as `*int` which would force the user to initialize his struct in several lines such as: 
```
    limit = 10
    query: &TasksQuery{
        Limit: *limit,
    }
```
And I think for this specific case when `limit = 0` there returns an empty return array so I'm not sure that the inconvenience is worth it

.code-samples.meilisearch.yaml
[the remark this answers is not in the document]
brunoocasali: Yeah, it could work but not for every case unfortunately

PR #314, opened 2022-06-30
Apply documents changes

alallema:
# Pull Request

Related to: https://github.com/meilisearch/integration-guides/issues/205

## What does this PR do?

Breaking because enforces the users to use Meilisearch v0.28.0

- `GET /documents/:uid`, `GET /documents` Add the possibility to reduce the body payload by using a query parameter called `fields` (previously called `attributesToRetrieve`), check the issue and the spec for the entire behavior.
- The documents objects in now return wrap in a `results`
- [x] Adapt DocumentsQuery naming following all the routes
- [x] Fix tests

### DIFF
--- a/index_documents.go
+++ b/index_documents.go
@@ -30,7 +30,7 @@ func (i Index) GetDocument(identifier string, documentPtr interface{}) error {
 	return nil
 }
 
-func (i Index) GetDocuments(request *DocumentsRequest, resp interface{}) error {
+func (i Index) GetDocuments(request *DocumentsQuery, resp *DocumentsResult) error {
 	req := internalRequest{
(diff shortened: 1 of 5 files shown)


index_documents.go
```diff
@@ -40,14 +40,14 @@ func (i Index) GetDocuments(request *DocumentsRequest, resp interface{}) error {
 		acceptedStatusCodes: []int{http.StatusOK},
 		functionName:        "GetDocuments",
 	}
-	if request.Limit != 0 {
+	if request != nil && request.Limit != 0 {
```
brunoocasali, line 43: I think we can use the same idea here:

```go
if request != nil {
  ...
}
```

brunoocasali (approved):
LGTM! 🎉 🦴

PR #315, opened 2022-06-30
Apply tenant changes

alallema:
as per the [spec](https://github.com/meilisearch/specifications/pull/148/files#diff-7b1a49daa5acb02c131aff3eb5fabb81c7e7c3d1b95adb60c4bf70cadcb488d4)

- [x] apiKeyPrefix claim is now named apiKeyUid and expects the uid of the signing API key as a value.

### DIFF
--- a/client.go
+++ b/client.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"regexp"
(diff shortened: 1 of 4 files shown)


brunoocasali (approved):
Good job! 🎉

PR #321, opened 2022-07-07
Refactoring Task ressource

alallema:
# Pull Request

## What does this PR do?

The idea of this PR is to create two separate structures to handle the `Task` as a resource: the return of the `GET /tasks` route and `TaskInfo` the return of all async routes.

### DIFF
--- a/types.go
+++ b/types.go
@@ -103,12 +103,29 @@ const (
 	TaskStatusFailed TaskStatus = "failed"
 )
 
-// Task indicate information about a task is returned for asynchronous method
+// Task indicate information about a task ressource
 //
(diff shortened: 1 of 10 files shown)


types.go
```diff
@@ -120,6 +120,23 @@ type Task struct {
 	Details    Details             `json:"details,omitempty"`
 }
 
+
+// TaskInfo indicate Information of the regarding a task returned by asynchronous method
```
brunoocasali, line 124:
```suggestion
// TaskInfo indicates information regarding a task returned by an asynchronous method
```

brunoocasali (approved):
🎉

PR #342, opened 2022-08-22
Create key: Add ability to specify the `Uid` and `Name` value

dbolkensteyn:
`type Key struct {}` and `type KeyParsed struct {}` both have a `Key string` field.

However, `KeyParsed`'s one is never filled and thus the server always will generate a random key, even if a value was specified in the call to `client.CreateKey(&meilisearch.Key{Key: "..."}}`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -385,7 +385,7 @@ func (c *Client) GenerateTenantToken(APIKeyUID string, SearchRules map[string]in
 // and transform the Key structure into a KeyParsed structure to send the time format
 // managed by Meilisearch
 func convertKeyToParsedKey(key Key) (resp KeyParsed) {
-	resp = KeyParsed{Description: key.Description, Actions: key.Actions, Indexes: key.Indexes}
+	resp = KeyParsed{Description: key.Description, Key: key.Key, Actions: key.Actions, Indexes: key.Indexes}
 
 	// Convert time.Time to *string to feat the exact ISO-8601
 	// format of Meilisearch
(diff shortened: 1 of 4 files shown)


alallema (commented):
Hi @dbolkensteyn,
Thanks for your PR. Indeed, the key is not specified because `convertKeyToParsedKey()` is called only by the `CreateKey()` method, and no matter what key is passed to that method, it will create a new key. So yes it will always return a random new key.

dbolkensteyn:
Oh right, `key` is not present as a parameter in the REST API apparently: https://docs.meilisearch.com/reference/api/keys.html#create-a-key

That's a bummer since it makes it impossible to create a predictable search key.

dbolkensteyn:
Hey @alallema, could you please review this once more?

They `Key` parameter isn't part of the REST API, but `UID` is. Since `Key` is derived from the master key and `UID`, it becomes predictable, thus solving my original issue.

This PR now sends both `UID` and `Name` (which seemed to have been forgotten) to the REST API. I've also updated the corresponding unit test.

alallema (approved):
Thanks, @dbolkensteyn for this PR and for contributing to `Meilisearch` ❤️ 
Good catch and at the same time you fix the fact that the `Name` of the Key wasn't created 🔥 
LGTM!

alallema:
bors merge

dbolkensteyn:
Awesome thank you so much @alallema

PR #349, opened 2022-09-02
Add tests to wildcarded action keys

brunoocasali:
Ensure support to keys with wildcarded actions.

- `actions` field during key creation now accepts wildcards on actions. For example, `indexes.*` provides rights to `indexes.create`, `indexes.get`,`indexes.delete`, `indexes.delete`, and `indexes.update`.


alallema (approved):
LGTM! Thanks @brunoocasali 🌮

PR #377, opened 2022-10-26
add SearchRaw to return raw JSON from search

### LINKED ISSUE #337: a `json.RawMessage` search response's hit fields
Previous I'm using a es client to search.

This es client has a feature that it's search response's hit field has s `Source` field with `json.RawMessage` field, contain the raw bytes from API.

And so I can unmarshal it to the struct I'm expecting.

It's SDK api look like this:

```golang
		result, err := es.search(...)
		if err != nil {
			return errgo.Wrap(err, "search")
		}

		for i, hit := range result.Hits.Hits {
			var source = Subject{}
// and hit.Source is a json.RawMesssage, So I can unmarshal it here.
			err = json.Unmarshal(hit.Source, &source)
			if err != nil {
				return err
			}
```
(it's source https://github.com/olivere/elastic/blob/e5be2e003d85f8b9bf2c97662e34c6a857f60a61/search.go#L779)

But, currently, meilisearch's `SearchResponse.Hits` is a `[]interface{}`, so I'll need to do many type case or marshal to json bytes then unmarshal to back.

Hope there could be some API improvement on this.

Or I hope it's possible I can get the raw response bytes somehow so I can unmarshal from the response directly.

ginglis13:
adds a method SearchRaw with the same signature as Search except it returns json.RawMessage

# Pull Request

## Related issue
Fixes #337

## What does this PR do?
- add `SearchIndexRaw` to index interface as a method to return a `json.RawMessage` from a search
- adds unit test for the method. The test has a `want` of type `SearchResponse`; this is because to verify the method, the test is unmarshalling the raw response returned into a `SearchResponse`

## PR checklist


index_search.go
```diff
@@ -11,74 +12,106 @@ const (
...
+		contentType:         contentTypeJSON,
+		withRequest:         searchPostRequestParams,
+		withResponse:        resp,
+		acceptedStatusCodes: []int{http.StatusOK},
+		functionName:        "Search",
+	}
+
+	if err := i.client.executeRequest(req); err != nil {
+		return nil, err
+	}
```
alallema, line 62: Maybe this method can call directly `SearchRaw` and Unmarshal the result in `SearchResponse` to avoid duplicate code, WDYT?

alallema (commented):
Thanks, @ginglis13 for this PR it will be a really great addition 🚀 
I just made a suggestion 😊

alallema:
Hi @ginglis13,
FYI, I will not be available from tonight until November 7th. Also, I already put the label `hacktoberfest-accepted` on your PR in case I won't have time to merge it before the end of the hacktoberfest 🥳

index_search.go
ginglis13: Ah great point... I was essentially doing that in the unit test but not the actual implementation 😅 I'll take your suggestion and make that change

trim21:
Hi, have you consider just change `Hits` field to `json.RawMessage` to use other fields?

```golang
// SearchResponseRaw is the response body for search method without decoding Hits.
type SearchResponseRaw struct {
	Hits               json.RawMessage `json:"hits"`
	EstimatedTotalHits int64           `json:"estimatedTotalHits"`
	Offset             int64           `json:"offset"`
	Limit              int64           `json:"limit"`
	ProcessingTimeMs   int64           `json:"processingTimeMs"`
	Query              string          `json:"query"`
	FacetDistribution  json.RawMessage `json:"facetDistribution,omitempty"`
}
```

alallema:
> Hi, have you consider just change Hits field to json.RawMessage to use other fields? So if want to get EstimatedTotalHits, so no need to unmarshal SearchResponse again.

Yes, but for me, it was simpler to propose a method that returns either a ready object with the existing structure in the SDK or returns a complete object in `json. RawMessage` so that the user has the choice to receive either a complete raw object or a pre-formatted object.

alallema (approved):
Thanks a lot @ginglis13 for this PR and the great addition to this SDK 🎉 
LGTM! 🔥 
Thanks again for contributing with Meilisearch :heart:
If you are participating in [Hacktoberfest](https://blog.meilisearch.com/contribute-hacktoberfest-2022/), and you would like to receive some swag from Meilisearch too, please complete [this form](https://meilisearch.typeform.com/to/idt5qugY).

alallema:
bors merge

PR #395, opened 2022-12-07
Add cancel task api for v0.30.0

alallema:
Add the cancel task api see spec:  https://github.com/meilisearch/specifications/pull/195

- [x] Update `Task` structure
  - [x] Add `canceledBy`
- [x] Update `TaskInfo` structure
  - [x] All parameters `Status`, `Type`, `TaskUID`, `IndexUID` and `EnqueuedAt` are no longer optional.
- [x] Implement `CancelTasks()` in the client
  - params: `TasksQuery`
  - returns: `TaskInfo`
- [x] add new `Tasks.Details` fields:
   - [x] matchedTasks
   - [x] canceledTasks
   - [x] originalFilters
- [x] Tests cancelation on query parameters
- [x] Test on new details

### DIFF
--- a/client.go
+++ b/client.go
@@ -54,7 +54,7 @@ type ClientInterface interface {
 	IsHealthy() bool
 	GetTask(taskUID int64) (resp *Task, err error)
 	GetTasks(param *TasksQuery) (resp *TaskResult, err error)
-	CancelTasks(param *TasksQuery) (resp *TaskInfo, err error)
+	CancelTasks(param *CancelTasksQuery) (resp *TaskInfo, err error)
 	WaitForTask(taskUID int64, options ...WaitParams) (*Task, error)
 	GenerateTenantToken(APIKeyUID string, searchRules map[string]interface{}, options *TenantTokenOptions) (resp string, err error)
 }
@@ -286,7 +286,7 @@ func (c *Client) GetTasks(param *TasksQuery) (resp *TaskResult, err error) {
 	return resp, nil
 }
 
-func (c *Client) CancelTasks(param *TasksQuery) (resp *TaskInfo, err error) {
+func (c *Client) CancelTasks(param *CancelTasksQuery) (resp *TaskInfo, err error) {
 	resp = &TaskInfo{}
 	req := internalRequest{
 		endpoint:            "/tasks/cancel",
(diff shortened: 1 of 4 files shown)


brunoocasali (approved):

client.go
```diff
@@ -285,6 +286,26 @@ func (c *Client) GetTasks(param *TasksQuery) (resp *TaskResult, err error) {
 	return resp, nil
 }
 
+func (c *Client) CancelTasks(param *TasksQuery) (resp *TaskInfo, err error) {
```
brunoocasali, line 289: You cannot use `TasksQuery` as a parameter here, because this method does not accept `limit` and `from`

image

brunoocasali (changes requested):

client.go
brunoocasali: I also updated to **- Guarantee all the filters available in the `GET /tasks` but `from` and `limit` are available.** I hope it's clearer

alallema: It will not send `TaskQuery` directly:
```go
	if param != nil {
		encodeTasksQuery(param, &req)
	}
	if err := c.executeRequest(req); err != nil {
		return nil, err
	}
```
But it will only encode the parameters set on the structure see `encodeTasksQuery` method

brunoocasali: I know that, `TaskQuery` contains two fields called `limit` and `from`, and they are not allowed in the meilisearch request `POST /tasks/cancel`. But if you share the type between this method and `get_tasks` for example, you'll support it and this is not correct :)

If you didn't get my explanation, please try to send a request to `POST /tasks/cancel` with `limit` and `from` body parameters, and see what happens, or read the spec again 😃

alallema: I understood very well what you meant, and it is I who wrote this spec 😄
I know that the `TaskQuery` structure is every fields possible for the queries made on the `task` route, but I think that precisely the fact that the structure itself is not sent, we avoid having to create code duplication by creating three specific structures for each query.

types.go
```diff
@@ -161,6 +161,18 @@ type TasksQuery struct {
...
+type CancelTasksQuery struct {
+	UIDS             []int64
+	IndexUIDS        []string
+	Statuses         []string
+	Types            []string
+	BeforeEnqueuedAt time.Time
+	AfterEnqueuedAt  time.Time
+	BeforeStartedAt  time.Time
+	AfterStartedAt   time.Time
+}
```
brunoocasali, line 174:
```suggestion
type CancelTasksQuery struct {
	UIDS             []int64
	IndexUIDS        []string
	Statuses         []string
	Types            []string
	BeforeEnqueuedAt time.Time
	AfterEnqueuedAt  time.Time
	BeforeStartedAt  time.Time
	AfterStartedAt   time.Time
	BeforeFinishedAt time.Time
  AfterFinishedAt time.Time
}
```

client.go
```diff
@@ -298,7 +298,17 @@ func (c *Client) CancelTasks(param *TasksQuery) (resp *TaskInfo, err error) {
...
+			UIDS:             param.UIDS,
+			IndexUIDS:        param.IndexUIDS,
+			Statuses:         param.Statuses,
+			Types:            param.Types,
+			BeforeEnqueuedAt: param.BeforeEnqueuedAt,
+			AfterEnqueuedAt:  param.AfterEnqueuedAt,
+			BeforeStartedAt:  param.BeforeStartedAt,
+			AfterStartedAt:   param.AfterStartedAt,
+		}
+		encodeTasksQuery(paramToSend, &req)
```
brunoocasali, line 311: Can you make this function `encodeTasksQuery` support multiple sets of params?

types.go
alallema: I didn't add those two filters because they didn't make sense, and they are not in the parameters available in the documentation. Are you sure it's a good think to add it?

client.go
alallema: Not really, I could put `interface{}` as a parameter instead of a typed struct, but I wouldn't have access to the fields easily.

brunoocasali (approved):
LGTM!

PR #403, opened 2022-12-22
Add yaml linter test in ci

bidoubiwa:
To ensure no yaml file are broken, I added a linter test on the yaml files. 
This ensures they are correctly formated and correctly linted.

To test it locally you need to use the [`yamllint` tool](https://github.com/adrienverge/yamllint) and then run it like that:

```
yamllint .
```

Example of output

Screenshot 2022-12-22 at 14 22 38

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -290,7 +290,7 @@ getting_started_typo_tolerance: |-
       OneTypo: 4,
     },
   })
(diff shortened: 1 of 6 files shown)


brunoocasali (approved):
Thanks a lot @bidoubiwa let's make it a standard in our SDKs :)

CC: @alallema I'll merge this since the code-samples are breaking in the docs site.

bors merge

PR #407, opened 2023-01-25
`Limit` redefinition in `DocumentsQuery`

### LINKED ISSUE #331: Find a better way to define `Limit` parameter in all `QueryParameter` structs
**Description**
In all the methods that allow sending a query parameter as struct like:
- TasksQuery
- DocumentsQuery
- IndexesQuery
- KeysQuery
We check that the parameter limit exists:
```
    if param.Limit != 0 {
```
Yet, Meilisearch allows setting the limit parameter to 0, which is not possible currently with this SDK because the `Limit` is define as an `int64`
```
    Limit   int64   `json:"limit"`
```

The simple solution could be to define the `Limit` as an `*int` but in this case, it would force the user to initialize his struct in several lines such as:
```
    limit = 10
    query: &TasksQuery{
        Limit: *limit,
    }
```
instead of simply doing it as it is today:
```
    query: &TasksQuery{
        Limit: 10,
    }
```

It would be nice to find a solution that does not complicate the use for the user and allow at the same time to send the value 0

Azanul:
# Pull Request

## Related issue
Fixes #331 

## What does this PR do?
- Adds support for sending 0 (zero) `Limit` when querying documents
- Creates a helper function for converting `int64` to `*int64`
- Adapts respective tests
- Regenerates easyjson file

## PR checklist

### DIFF
--- a/index_documents.go
+++ b/index_documents.go
@@ -45,8 +45,8 @@ func (i Index) GetDocuments(request *DocumentsQuery, resp *DocumentsResult) erro
 		functionName:        "GetDocuments",
 	}
 	if request != nil {
-		if request.Limit != 0 {
-			req.withQueryParams["limit"] = strconv.FormatInt(request.Limit, 10)
+		if *request.Limit != 0 {
+			req.withQueryParams["limit"] = strconv.FormatInt(*request.Limit, 10)
 		}
 		if request.Offset != 0 {
 			req.withQueryParams["offset"] = strconv.FormatInt(request.Offset, 10)
@@ -413,4 +413,9 @@ func (i Index) DeleteAllDocuments() (resp *TaskInfo, err error) {
 		return nil, err
 	}
 	return resp, nil
+}
+
+// Helper function to create a pointer to int64
+func int64p(i int64) *int64 {
+	return &i
 }
--- a/types.go
+++ b/types.go
@@ -340,7 +340,7 @@ type DocumentQuery struct {
 // DocumentsQuery is the request body for list documents method
 type DocumentsQuery struct {
 	Offset int64    `json:"offset,omitempty"`
-	Limit  int64    `json:"limit,omitempty"`
+	Limit  *int64   `json:"limit,omitempty"`
 	Fields []string `json:"fields,omitempty"`
 }
 
--- a/index_documents_test.go
+++ b/index_documents_test.go
@@ -205,7 +205,7 @@ func TestIndex_AddDocuments(t *testing.T) {
 			testWaitForTask(t, i, gotResp)
 			var documents DocumentsResult
 			err = i.GetDocuments(&DocumentsQuery{
-				Limit: 3,
+				Limit: int64p(3),
 			}, &documents)
 			require.NoError(t, err)
 			require.Equal(t, tt.resp.documentsRes, documents)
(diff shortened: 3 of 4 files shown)


brunoocasali:
@Azanul is this PR finished? Can we review it?

Azanul:
@brunoocasali I don't know if the discussion in the linked issue was conclusive or not. If so then yes.

alallema:
Hi @Azanul,
> I don't know if the discussion in the linked issue was conclusive or not.

Indeed, the discussion was not concluded, but as said in my comment, I would lean toward the NoLimit global variable option. Again I apologize for the delay 😊

Azanul:
@alallema @aldy505  I've updated the implementation and the tests according to global var approach.

Azanul:
@brunoocasali Now it's ready for review

index_documents.go
```diff
@@ -15,6 +15,8 @@ import (
 	"strings"
 )
 
+var NoLimit int64 = math.MinInt64
```
aldy505, line 18: We should provide comment (as documentation for users) to what this variable means, and when to use it.

aldy505 (changes requested):

index_documents.go
Azanul: check it out

aldy505: I'll give a suggestion

types.go
```diff
@@ -350,7 +350,7 @@ type DocumentQuery struct {
 // DocumentsQuery is the request body for list documents method
 type DocumentsQuery struct {
 	Offset int64    `json:"offset,omitempty"`
-	Limit  int64    `json:"limit,omitempty"`
+	Limit  int64   `json:"limit,omitempty"`
```
aldy505, line 353:
```suggestion
	// Limit sets the numbers of returned documents from the
	// search queries. The limit can be zero for explicitly
	// return nothing. For retrieving all 
	// returned documents, use NoLimit. 
	Limit  int64   `json:"limit,omitempty"`
```

index_documents.go
```diff
@@ -15,6 +15,9 @@ import (
 	"strings"
 )
 
+// Variable for setting no limit on number of documents to fetch
```
aldy505, line 18:
```suggestion
// NoLimit should be used to set Limit field value
// on DocumentsQuery struct to return all documents.
```

aldy505 (changes requested):
Here are my suggested changes. You can directly apply it by pressing the "commit suggested changes", otherwise you can retype and commit it yourself.

types.go
Azanul: I don't think adding a code comment to every single attribute is a good practice. The code is going to look bloated.

aldy505: No one reads the code. But everyone will benefits from having a documentation on their IDE/text-editor without having to go to the documentation. All you need to do is just hover at the struct/field name.

Take `http.Client` documentation for example: https://pkg.go.dev/net/http@go1.20.4#Client -- they have comments as documentation on everything. It helps the users understand what that is. If you look at the source code (https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/net/http/client.go;l=30-107), yes it's a bit messy thanks to lots and lots of documentation. But as the user who uses the library, it's helpful.

Azanul: Great example! Thanks a lot for the reference.

aldy505 (approved):

brunoocasali:
> @alallema Seems like newly added Filter code doesn't respect Limit. Do you want me to update respectively.

Yes that would be nice @Azanul :)

Azanul:
@alallema @brunoocasali @aldy505 Do any of you know how I can adapt the current approach for adding 0 `Limit` support to work with the existing `Filter` implementation.
I can go back to pointer approach or I'll have to make changes to `Filter` implementation.

alallema:
Hi @Azanul,
Thank you so much for all your hard work on this PR.
Unfortunately, we are reducing the support on this repo, and as this issue is not essential and will have to be postponed for Task, Keys etc. I propose to close it.
Thanks again for your involvement and participation ❤️

PR #408, opened 2023-01-26
Implement update methods CSV & NdJson

### LINKED ISSUE #215: NDJSON/CSV methods to add and update documents

Azanul:
# Pull Request

## Related issue
Fixes #215 

## What does this PR do?
- [X] Add `UpdateDocumentsInBatches` to `IndexInterface`
- [X] Implement `UpdateDocumentsCsv`
- [X] Implement `UpdateDocumentsCsvInBatches`
- [X] Implement `UpdateDocumentsNdjson`
- [X] Implement `UpdateDocumentsNdjsonInBatches`
- [X] Implement respective tests
- [X] Implement `saveDocuments` function to reduce code duplication

## PR checklist

### DIFF
--- a/index_documents.go
+++ b/index_documents.go
@@ -372,7 +372,7 @@ func (i Index) UpdateDocumentsInBatches(documentsPtr interface{}, batchSize int,
 }
 
 func (i Index) UpdateDocumentsCsv(documents []byte, primaryKey ...string) (resp *TaskInfo, err error) {
-	return i.updateDocuments(documents, contentTypeJSON, primaryKey...)
+	return i.updateDocuments(documents, contentTypeCSV, primaryKey...)
 }
 
 func (i Index) UpdateDocumentsCsvFromReader(documents io.Reader, primaryKey ...string) (resp *TaskInfo, err error) {
--- a/index_documents.go
+++ b/index_documents.go
@@ -143,7 +143,7 @@ func (i Index) AddDocumentsCsvInBatches(documents []byte, batchSize int, primary
 	return i.AddDocumentsCsvFromReaderInBatches(bytes.NewReader(documents), batchSize, primaryKey...)
 }
 
-func sendCsvRecords(documentsCsvFunc func, records [][]string, primaryKey ...string) (*TaskInfo, error) {
+func sendCsvRecords(documentsCsvFunc func(recs []byte, pk ...string) (resp *TaskInfo, err error), records [][]string, primaryKey ...string) (*TaskInfo, error) {
 	b := new(bytes.Buffer)
 	w := csv.NewWriter(b)
 	w.UseCRLF = true
@@ -212,7 +212,7 @@ func (i Index) AddDocumentsCsvFromReaderInBatches(documents io.Reader, batchSize
 
 	// Send remaining records as the last batch if there is any
 	if len(records) > 0 {
-		resp, err := sendCsvRecords(records)
+		resp, err := sendCsvRecords(i.AddDocumentsCsv, records, primaryKey...)
 		if err != nil {
 			return nil, err
 		}
@@ -432,7 +432,7 @@ func (i Index) updateDocumentsCsvFromReaderInBatches(documents io.Reader, batchS
 
 	// Send remaining records as the last batch if there is any
 	if len(records) > 0 {
-		resp, err := sendCsvRecords(records)
+		resp, err := sendCsvRecords(i.UpdateDocumentsCsv, records, primaryKey...)
 		if err != nil {
 			return nil, err
 		}
--- a/index_documents.go
+++ b/index_documents.go
@@ -461,6 +461,78 @@ func (i Index) UpdateDocumentsNdjson(documents []byte, primaryKey ...string) (re
 	return i.updateDocuments(documents, contentTypeNDJSON, primaryKey...)
 }
 
+func (i Index) UpdateDocumentsNdjsonInBatches(documents []byte, batchsize int, primaryKey ...string) (resp []TaskInfo, err error) {
+	return i.updateDocumentsNdjsonFromReaderInBatches(bytes.NewReader(documents), batchsize, primaryKey...)
+}
+
+func (i Index) updateDocumentsNdjsonFromReaderInBatches(documents io.Reader, batchSize int, primaryKey ...string) (resp []TaskInfo, err error) {
+	// NDJSON files supposed to contain a valid JSON document in each line, so
+	// it's safe to split by lines.
+	// Lines are read and sent continuously to avoid reading all content into
+	// memory. However, this means that only part of the documents might be
+	// added successfully.
+
+	sendNdjsonLines := func(lines []string) (*TaskInfo, error) {
+		b := new(bytes.Buffer)
+		for _, line := range lines {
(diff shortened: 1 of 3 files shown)


index_documents.go
```diff
@@ -366,6 +370,169 @@ func (i Index) UpdateDocumentsInBatches(documentsPtr interface{}, batchSize int,
...
+	sendCsvRecords := func(records [][]string) (*TaskInfo, error) {
+		b := new(bytes.Buffer)
+		w := csv.NewWriter(b)
+		w.UseCRLF = true // Keep output RFC 4180 compliant
+		err := w.WriteAll(records)
+		if err != nil {
+			return nil, fmt.Errorf("could not write CSV records: %w", err)
+		}
+
+		resp, err := i.UpdateDocumentsCsv(b.Bytes(), primaryKey...)
```
alallema, line 406: Except for this line, this method is the exact same as `AddDocumentsNdjsonFromReaderInBatches` maybe we can find a way to re-used the code that already existing than duplicated it? WDYT?

index_documents.go
```diff
@@ -366,6 +370,169 @@ func (i Index) UpdateDocumentsInBatches(documentsPtr interface{}, batchSize int,
...
+			if err != nil {
+				return nil, fmt.Errorf("Could not write NDJSON line: %w", err)
+			}
+			err = b.WriteByte('\n')
+			if err != nil {
+				return nil, fmt.Errorf("Could not write NDJSON line: %w", err)
+			}
+		}
+
+		resp, err := i.UpdateDocumentsNdjson(b.Bytes(), primaryKey...)
```
alallema, line 488: Same for this method maybe a trick exist to avoid the duplication of this piece of code

alallema (commented):
Thanks @Azanul  for this PR!
I have added a few comments. Also every time you create a method or modify something, you need to create the associated tests to check it.
Don't forget to put your PR as `Ready for review` when it is ready and you can request me as a reviewer when you need a review 😊 
Screenshot 2023-02-08 at 15 03 55

index_documents.go
Azanul: Sure, I think a higher order function will resolve this.

Azanul: > Except for this line, this method is the exact same as `AddDocumentsNdjsonFromReaderInBatches` maybe we can find a way to re-used the code that already existing than duplicated it? WDYT?

I hope you mean `AddDocumentsCsvFromReaderInBatches`

index_documents.go
Azanul: check it out, if you approve this approach, I'll use the same for other methods too where this pattern is occurring.

alallema: It's better, but the problem is that almost all the method is duplicated from line 408 to 452 is possible to group all the duplicated code?

Azanul: How about using this same approach to implement a template design pattern?

Azanul: Check it out. Also, should I be doing the same for from reader functions?

index_documents.go
```diff
@@ -70,26 +39,30 @@ func (i Index) addDocuments(documentsPtr interface{}, contentType string, primar
...
 		withResponse:        resp,
 		acceptedStatusCodes: []int{http.StatusAccepted},
-		functionName:        "AddDocuments",
+		functionName:        function + "Documents",
 	}
 	if err = i.client.executeRequest(req); err != nil {
 		return nil, err
 	}
 	return resp, nil
 }
```
alallema, line 63: Sorry, I realized that I may not have been clear. I was not talking about refactoring the code of all the functions; for me this method (`saveDocuments`) only makes the understanding of the code more complex
I was talking about lines 367 to 398 (previously 408 to 452) which are always duplicated and could be put in the same function.

Azanul: Check it out if that's what you meant

index_documents.go
```diff
@@ -70,80 +39,30 @@ func (i Index) addDocuments(documentsPtr interface{}, contentType string, primar
...
```
alallema, line 63: This method isn't needed you can call `addDocuments()` or `updateDocuments()` instead

index_documents.go
alallema: Sorry for the delay.
That's right, except you don't have to rewrite the `addDocuments()` method, we don't need the `saveDocument()`.

alallema (approved):
Thank you a lot @Azanul for this PR and all the modifications ✨ 
This will be a great enhancement 🔥 
LGTM! 🎉

PR #422, opened 2023-03-22
Add multiple searches

alallema:
Now Meilisearch can perform multiple searches in one single HTTP request.
A new method has been added to the [specifications](https://github.com/meilisearch/specifications/pull/225):
`MultiSearch(queries MultiSearchRequest) (*MultiSearchResponse, error)`

The method takes `[]SearchRequest` as argument:
```go
	m := []meilisearch.SearchRequest{}
	m = append(m, meilisearch.SearchRequest{
		IndexUid: "movies",
		Query: "wonder",
	})
	client.MultiSearch(m)
```

SDK requirements: https://github.com/meilisearch/integration-guides/issues/251

### DIFF
--- a/client.go
+++ b/client.go
@@ -42,6 +42,7 @@ type ClientInterface interface {
 	CreateIndex(config *IndexConfig) (resp *TaskInfo, err error)
 	DeleteIndex(uid string) (resp *TaskInfo, err error)
 	CreateKey(request *Key) (resp *Key, err error)
+	MultiSearch(queries *MultiSearchRequest) (*MultiSearchResponse, error)
 	GetKey(identifier string) (resp *Key, err error)
 	GetKeys(param *KeysQuery) (resp *KeysResults, err error)
 	UpdateKey(keyOrUID string, request *Key) (resp *Key, err error)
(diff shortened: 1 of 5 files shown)


bidoubiwa:
Only missing the label

client.go
```diff
@@ -252,6 +253,35 @@ func (c *Client) CreateDump() (resp *TaskInfo, err error) {
 	return resp, nil
 }
 
+func (c *Client) MultiSearch(queries []SearchRequest) (*MultiSearchResponse, error) {
```
bidoubiwa, line 256: The user should provide `{ queries:  []SearchRequest }` not just ` []SearchRequest`

types.go
```diff
@@ -328,6 +330,11 @@ type SearchResponse struct {
 	HitsPerPage        int64         `json:"hitsPerPage,omitempty"`
 	Page               int64         `json:"page,omitempty"`
 	TotalPages         int64         `json:"totalPages,omitempty"`
+	IndexUid           string        `json:"indexUid,omitempty"`
```
bidoubiwa, line 333: This is not the case when performing a simple `search`. `MultiSearchResponse` should be a union of `SearchResponse` and

```
	IndexUid           string        `json:"indexUid,omitempty"`
```

index_search.go
```diff
@@ -65,11 +65,14 @@ func (i Index) Search(query string, request *SearchRequest) (*SearchResponse, er
...
 func searchPostRequestParams(query string, request *SearchRequest) map[string]interface{} {
-	params := make(map[string]interface{}, 14)
+	params := make(map[string]interface{}, 16)
 
 	if !request.PlaceholderSearch {
 		params["q"] = query
 	}
+	if request.IndexUid != "" {
+		params["indexUid"] = request.IndexUid
+	}
```
bidoubiwa, line 75: indexUid does not exist on simple search I think

client.go
alallema: Yes, but it is much less annoying for the user to send a search request table than to create a map to put his table in. So I do it for them

bidoubiwa: It's up to you! The idea of adding it in the object is so that if in the future their are other fields than `queries` you don't have to do a breaking change

types.go
alallema: That is why I used `omitempty` which meant that go will not create the fields if it doesn't exist.
What do you mean by Union? The type `Union` in golang or something like:
```
type MultiSearchResponse struct {
	Results []MultiSearchUnionResponse `json:"results"`
}

type MultiSearchUnionResponse struct {
	SearchResponse SearchResponse{}
	IndexUid string
}
```
?
Because if I use the second option, you will have to do something like:
```
myResponse.Results[0].SearchResponse.Limit
```
Instead of:
```
myResponse.Results[0].Limit
```
Like it is in the response API

index_search.go
alallema: Yes, that why this condition is made for if there is any specified `IndexUid` it will not create the field

bidoubiwa: Yes but is this not lying to the user? You tell them that `indexUid` is an accepted query parameter when it is not.

alallema: Yes, it because I used the same struct `SearchResponse` than for the `Search()` and for `MultiSearch`

bidoubiwa (approved):
LGTM 🔥

PR #423, opened 2023-03-22
Add `FacetStats` field in `SearchResponse`

alallema:
As per [the specification](https://github.com/meilisearch/specifications/pull/224)
SDK requirements: https://github.com/meilisearch/integration-guides/issues/251

A new `facetStats` field is returned when `facets` is used in the search parameters and contains numerical values. The min and max numeric values of these facets will be returned.

@brunoocasali I added a test in the case tests to check that the returned types were well declared

### DIFF
--- a/index_search_test.go
+++ b/index_search_test.go
@@ -524,6 +524,45 @@ func TestIndex_SearchFacets(t *testing.T) {
 					}),
 			},
 		},
+		{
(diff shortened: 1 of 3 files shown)


bidoubiwa (approved):
LGTM 🔥


Only missing the label

PR #439, opened 2023-05-23
Adds filter to the parameters for GetDocuments - Meilisearch v1.2

alallema:
As per the specification: https://github.com/meilisearch/specifications/pull/234

- Add ability to `GetDocuments` with a filter
- Add the `filter` field on the `DocumentsQuery` struct
   - The `filter` is an `interface{}` which allows you to send different types of filters (`string`, `[]string`, `[]interface{}{[]string{}}`, ...)

The `filter` field works precisely like the `filter` field used on the `search` method. See [the docs on how to use filters](https://www.meilisearch.com/docs/learn/advanced/filtering#filter-basics).

### DIFF
--- a/error_test.go
+++ b/error_test.go
@@ -20,7 +20,9 @@ func TestError_VersionErrorHintMessage(t *testing.T) {
 		{
 			name: "VersionErrorHintMessageGetDocument",
 			args: args{
-				request: &internalRequest{},
+				request: &internalRequest{
+					functionName: "GetDocuments",
+				},
 				mockedError: &Error{
 					Endpoint:         "endpointForGetDocuments",
 					Method:           http.MethodPost,
@@ -41,7 +43,7 @@ func TestError_VersionErrorHintMessage(t *testing.T) {
 			err := VersionErrorHintMessage(tt.args.mockedError, tt.args.request)
 			require.Error(t, err)
 			fmt.Println(err)
-			require.Equal(t, tt.args.mockedError.Error()+". Hint: It might not be working because you're not up to date with the Meilisearch version that  call requires.", err.Error())
+			require.Equal(t, tt.args.mockedError.Error()+". Hint: It might not be working because you're not up to date with the Meilisearch version that "+tt.args.request.functionName+" call requires.", err.Error())
 		})
 	}
 }
--- a/error_test.go
+++ b/error_test.go
@@ -1,0 +1,47 @@
+package meilisearch
+
+import (
+	"fmt"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestError_VersionErrorHintMessage(t *testing.T) {
+	type args struct {
+		request     *internalRequest
+		mockedError error
+	}
+	tests := []struct {
+		name string
+		args args
+	}{
+		{
+			name: "VersionErrorHintMessageGetDocument",
+			args: args{
+				request: &internalRequest{},
+				mockedError: &Error{
+					Endpoint:         "endpointForGetDocuments",
+					Method:           http.MethodPost,
+					Function:         "GetDocuments",
+					RequestToString:  "empty request",
+					ResponseToString: "empty response",
+					MeilisearchApiError: meilisearchApiError{
+						Message: "empty Meilisearch message",
+					},
+					StatusCode: 1,
+					rawMessage: "empty message",
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := VersionErrorHintMessage(tt.args.mockedError, tt.args.request)
+			require.Error(t, err)
+			fmt.Println(err)
+			require.Equal(t, tt.args.mockedError.Error()+". Hint: It might not be working because you're not up to date with the Meilisearch version that  call requires.", err.Error())
+		})
+	}
+}
--- a/index_documents.go
+++ b/index_documents.go
@@ -184,7 +184,7 @@ func (i Index) GetDocuments(request *DocumentsQuery, resp *DocumentsResult) erro
 		acceptedStatusCodes: []int{http.StatusOK},
 		functionName:        "GetDocuments",
 	}
-	if request != nil && len(request.Filter) == 0 {
+	if request != nil && request.Filter == nil {
 		req.withQueryParams = map[string]string{}
 		if request.Limit != 0 {
 			req.withQueryParams["limit"] = strconv.FormatInt(request.Limit, 10)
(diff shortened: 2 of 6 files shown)


error_test.go
```diff
@@ -0,0 +1,47 @@
...
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := VersionErrorHintMessage(tt.args.mockedError, tt.args.request)
+			require.Error(t, err)
+			fmt.Println(err)
+			require.Equal(t, tt.args.mockedError.Error()+". Hint: It might not be working because you're not up to date with the Meilisearch version that  call requires.", err.Error())
```
brunoocasali, line 44:
```suggestion
			require.Equal(t, tt.args.mockedError.Error()+". Hint: It might not be working because maybe you're not up to date with the Meilisearch version that ${method} call requires.", err.Error())
```

You need to give the hint to the user including the name of the method!

index_documents.go
```diff
@@ -177,13 +177,15 @@ func (i Index) GetDocuments(request *DocumentsQuery, resp *DocumentsResult) erro
...
+		contentType:         contentTypeJSON,
 		withRequest:         nil,
 		withResponse:        resp,
-		withQueryParams:     map[string]string{},
+		withQueryParams:     nil,
 		acceptedStatusCodes: []int{http.StatusOK},
 		functionName:        "GetDocuments",
 	}
-	if request != nil {
+	if request != nil && len(request.Filter) == 0 {
```
brunoocasali, line 187: You did not defined a `else` clause, is that even possible to happen?

index_documents_test.go
```diff
@@ -1525,30 +1523,66 @@ func TestIndex_GetDocuments(t *testing.T) {
...
+				},
+			},
+			result: 3,
+		},
+		{
+			name: "TestIndexGetDocumentsWithMultipleFilter",
+			args: args{
+				client: defaultClient,
+				request: &DocumentsQuery{
+					Filter: []string{"tag = Tragedy", "book_id = 742"},
```
brunoocasali, line 1554: Will this be translated to "tag = Tragedy OR book_id = 742"?

error_test.go
alallema: Yes I forgot to mock the request too ... oops

index_documents.go
alallema: Yes, in the default case the `internalRequest` is set at the beginning of the method, avoid both if and is executed normally.

index_documents_test.go
alallema: `[]string{"tag = Tragedy", "book_id = 742"}` =  `"tag = Tragedy AND book_id = 742"`

```[]interface{}{[]string{"tag = Tragedy", "book_id = 123"}}``` =  `"tag = Tragedy OR book_id = 742"`

So I modify the type of filter `filter` from `[]string` to `interface{}` to allow the user to use both `string` as `[]string` as `interface{}`

Comment (brunoocasali) [reply index_documents_test.go:1568]: Sorry, but I didn't get it:

How this: `[]interface{}{[]string{"tag = Tragedy", "book_id = 123"}}` can be different than this `[]string{"tag = Tragedy", "book_id = 742"}`?

To me they should be different arrays: `[]interface{}{[]string{"tag = Tragedy", "book_id = 742"},[]string{...}` no?

File: index_documents_test.go [hunk 4]
```diff
@@ -1525,30 +1523,95 @@ func TestIndex_GetDocuments(t *testing.T) {
...
 
 			err := i.GetDocuments(tt.args.request, tt.args.resp)
 			require.NoError(t, err)
 			if tt.args.request != nil && tt.args.request.Limit != 0 {
 				require.Equal(t, tt.args.request.Limit, int64(len(tt.args.resp.Results)))
+			} else if tt.result != 0 {
+				require.Equal(t, tt.result, int64(len(tt.args.resp.Results)))
 			} else {
-				require.Equal(t, 6, len(tt.args.resp.Results))
+				require.Equal(t, 20, len(tt.args.resp.Results))
```
brunoocasali, line 1614: Why we just don't have a `require.Equal(t, tt.result, int64(len(tt.args.resp.Results)))` directly?

index_documents.go
brunoocasali: So you should not add the first `request != nil` :)

index_documents_test.go
alallema: This is not me; this is from the documentation guide about [filtering with array](https://www.meilisearch.com/docs/learn/advanced/filtering#creating-filter-expressions-with-arrays):
AND `["genres = horror", "director = 'Jordan Peele'"]`
OR `[["genres = horror", "genres = comedy"]]`

But in Golang you can do `[]string{[]string}` if you want to have an array of `[]string`, you should use `interface{}` or maybe there is another way but I don't know it.
But yes `[]interface{}{[]string{"tag = Tragedy", "book_id = 742"},[]string{...}` this should be also doable.

index_documents.go
alallema: > So you should not add the first request != nil

Not sure to understand well but in Go I had to check the pointer if unfortunately, I try to access a field of the request without checking if the request exists, I will encounter a panic.

But what I could do if you think is better is put to the `request != nil` in a upper level alone and the two other if inside.

index_documents_test.go
alallema: Because if the parameters or limit are not changed, there should be 20 results by default, so I haven't set any "result" in these test cases, and I'm verifying that there are indeed the expected default number of results.
I can modify it but I had to set the `results` is all this tests

brunoocasali: I see the point now. To be honest, I prefer seeing the duplicated values in each test case than understanding business logic in the tests. Tests are meant to be clear, some resources to back that POV:

- https://stackoverflow.com/a/129722/2649707
- https://stackoverflow.com/a/130038/2649707
- https://www.codewithjason.com/why-duplication-is-more-acceptable-in-tests/

Since you already have a setup where you write use cases by defining structs and then iterating over them, you have already reduced a lot of boilerplates. As you probably noted, the go tests differ from all the other SDKs.
I think you could reconsider the decision to keep those conditionals in the test and "duplicate" the setup in another issue.

brunoocasali (approved):

alallema:
> How this: []interface{}{[]string{"tag = Tragedy", "book_id = 123"}} can be different than this []string{"tag = Tragedy", "book_id = 742"}?

This `[]interface{}{[]string{"tag = Tragedy", "book_id = 123"}}` is equivalent to this `[["tag = Tragedy", "book_id = 123"]]` who is equivalent to `["tag = Tragedy" OR "book_id = 123"]`.

> To me they should be different arrays: []interface{}{[]string{"tag = Tragedy", "book_id = 742"},[]string{...} no?

While this `[]interface{}{[]string{"tag = Tragedy", "book_id = 742"},[]string{...}` is equivalent to `[["tag = Tragedy", "book_id = 123"], [...]]` and `[("tag = Tragedy" OR "book_id = 123") AND [...]]`

Not sure is more clear, but the main point to understand is that in Go we have to use `interface{}` to declare array of array of something so when we can easily write `[[ ... ]]` in Python or other languages in go is translated by `[]interface{}{[]string{}}`

PR #440, opened 2023-05-24
Add method DeleteDocumentsByFilter for Meilisearch v1.2

alallema:
As per the specification: https://github.com/meilisearch/specifications/pull/236

Added method `DeleteDocumentsByFilter`, this method takes an `interface{}` which allows you to send different types of filters (`string`, `[]string`, `[]interface{}{[]string{}}`, ...)

The `filter` field works precisely like the `filter` field used on the `search` method. See [the docs on how to use filters](https://www.meilisearch.com/docs/learn/advanced/filtering#filter-basics).

### DIFF
--- a/index_documents.go
+++ b/index_documents.go
@@ -526,6 +526,25 @@ func (i Index) DeleteDocuments(identifier []string) (resp *TaskInfo, err error) 
 	return resp, nil
 }
 
+func (i Index) DeleteDocumentsByFilter(filter interface{}) (resp *TaskInfo, err error) {
+	resp = &TaskInfo{}
+	req := internalRequest{
+		endpoint:    "/indexes/" + i.UID + "/documents/delete",
+		method:      http.MethodPost,
+		contentType: contentTypeJSON,
+		withRequest: map[string]interface{}{
+			"filter": filter,
+		},
+		withResponse:        resp,
+		acceptedStatusCodes: []int{http.StatusAccepted},
+		functionName:        "DeleteDocumentsByFilter",
+	}
+	if err := i.client.executeRequest(req); err != nil {
+		return nil, VersionErrorHintMessage(err, &req)
+	}
+	return resp, nil
+}
+
 func (i Index) DeleteAllDocuments() (resp *TaskInfo, err error) {
 	resp = &TaskInfo{}
 	req := internalRequest{
--- a/index_documents_test.go
+++ b/index_documents_test.go
@@ -1285,7 +1285,7 @@ func TestIndex_DeleteDocuments(t *testing.T) {
 		wantResp *TaskInfo
 	}{
 		{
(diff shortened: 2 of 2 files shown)


index_documents_test.go
```diff
@@ -1383,6 +1383,129 @@ func TestIndex_DeleteDocuments(t *testing.T) {
...
+		wantResp *TaskInfo
+	}{
+		{
+			name: "TestIndexDeleteDocumentsByFilterString",
+			args: args{
+				UID:            "1",
+				client:         defaultClient,
+				filterToApply:  []string{"book_id"},
+				filterToDelete: "book_id = 123",
+				documentsPtr: []docTestBooks{
```
brunoocasali, line 1406: What does ptr mean?

index_documents.go
brunoocasali, line 1: I don't know where I could comment that, but since the details have changed, wouldn't you have to update it in the types.go? https://github.com/meilisearch/meilisearch-go/blob/a0cbe0be6702d79004a81295168f1233d528c9f6/types.go#L191

brunoocasali: The original filters are going to be sent all the time https://github.com/meilisearch/meilisearch/releases/tag/v1.2.0-rc.2

brunoocasali (changes requested):

index_documents_test.go
alallema: These are auto-generated tests, I think it's for pointer.

index_documents.go
alallema: Need help to understand correctly. The `OriginalFilter` is already present in `Details` struct. Did you mean that this field will not be an option but will be present in every `Details` field all the time?

brunoocasali: Yes, when the type of the task is a `documentDeletion`.

index_documents.go
```diff
@@ -526,6 +526,25 @@ func (i Index) DeleteDocuments(identifier []string) (resp *TaskInfo, err error)
...
+			"filter": filter,
+		},
+		withResponse:        resp,
+		acceptedStatusCodes: []int{http.StatusAccepted},
+		functionName:        "DeleteDocumentsByFilter",
+	}
+	if err := i.client.executeRequest(req); err != nil {
+		return nil, VersionErrorHintMessage(err, &req)
+	}
+	return resp, nil
```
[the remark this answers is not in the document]
alallema: I can change it but all the others are formatted the same way

index_documents.go
alallema: Yes but `Details` is used for every type of `Task`

brunoocasali: Yes, but in your `Details` you had different fields in any case for every type of task. From document addition, to index settings configuration. It means this type serves multi purposes.

PR #459, opened 2023-08-03
Drop Golang 1.16, 1.17 support

alallema:
# Pull Request

According to the [end-of-life](https://endoflife.date/go) date for Golang versions, versions up to 1.18 are no longer supported. Knowing that Go generally only supports the last two versions

## What does this PR do?
- Removes support for Golang 1.16 and 1.17


alallema:
This PR should have worked, but it didn't and I don't have the time to work on it, so I'm shutting it down.

PR #481, opened 2023-09-03
Check if SearchRequest is nil

### LINKED ISSUE #480: If the SearchRequest object is nil, it returns an error instead of panic.
The index search methods takes a `SearchRequest` object as input, but they do not verify whether the object is nil or not. Consequently, a panic occurs when attempting to access the object.

luigibarbato:
# Pull Request

## Related issue
Fixes #480 

## What does this PR do?
In the current scenario, if a nil SearchRequest object is passed to the Index search methods, a panic occurs. This PR is intended to introduce an error handling to prevent unexpected panic.

## PR checklist

### DIFF
--- a/index_search.go
+++ b/index_search.go
@@ -2,6 +2,7 @@ package meilisearch
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
 )
 
@@ -12,14 +13,20 @@ const (
 	DefaultLimit int64 = 20
 )
 
+var ErrNoSearchRequest = errors.New("no search request provided")
+
 func (i Index) SearchRaw(query string, request *SearchRequest) (*json.RawMessage, error) {
-	resp := &json.RawMessage{}
+	if request == nil {
+		return nil, ErrNoSearchRequest
+	}
 
 	if request.Limit == 0 {
 		request.Limit = DefaultLimit
 	}
 
 	searchPostRequestParams := searchPostRequestParams(query, request)
+
+	resp := &json.RawMessage{}
 
 	req := internalRequest{
 		endpoint:            "/indexes/" + i.UID + "/search",
@@ -39,7 +46,9 @@ func (i Index) SearchRaw(query string, request *SearchRequest) (*json.RawMessage
 }
 
 func (i Index) Search(query string, request *SearchRequest) (*SearchResponse, error) {
-	resp := &SearchResponse{}
+	if request == nil {
+		return nil, ErrNoSearchRequest
+	}
 
 	if request.Limit == 0 {
 		request.Limit = DefaultLimit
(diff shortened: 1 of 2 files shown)


alallema (commented):
Thanks @luigibarbato for this PR,
I was wondering if it wouldn't be better to modify `searchPostRequestParams` method instead and return nil if `SearchRequest` is nil? That way, the user could still send nil in case of trouble and it would be more convenient, no?
WDYT?

luigibarbato:
> Thanks @luigibarbato for this PR, I was wondering if it wouldn't be better to modify the `searchPostRequestParams` method instead and return nil if `SearchRequest` is nil? That way, the user could still send nil in case of trouble and it would be more convenient, no? WDYT?

@alallema Thank you for your interest :)

Hmmm... [But we try to access the request before the `searchPostRequestParams` call](https://github.com/meilisearch/meilisearch-go/blob/c46c3f65c523a4e125a734f1a822e1e799ed4708/index_search.go#L18). 

So...  isn't the `SearchRequest` object mandatory for a success call? If not then trivially we could implement your suggestions by moving everything to the `searchPostRequestParams` method. But for what reason the object could be empty?

alallema:
Hi @luigibarbato,
Thanks for your answer 😃 

> Hmmm... [But we try to access the request before the searchPostRequestParams call](https://github.com/meilisearch/meilisearch-go/blob/c46c3f65c523a4e125a734f1a822e1e799ed4708/index_search.go#L18).

Yes indeed, but we could just change the condition to:
```
	if request != nil && request.Limit == 0 {
```
> But for what reason the object could be empty?

Because it could be convenient to write:
```
	client.Index("movies").Search("american ninja", nil)
```
Instead of:
```
	client.Index("movies").Search("american ninja", &meilisearch.SearchRequest{})
```
But really, I'm fine with both if you prefer to leave it as is.

luigibarbato:
> Hi @luigibarbato, Thanks for your answer 😃
> 
> > Hmmm... [But we try to access the request before the searchPostRequestParams call](https://github.com/meilisearch/meilisearch-go/blob/c46c3f65c523a4e125a734f1a822e1e799ed4708/index_search.go#L18).
> 
> Yes indeed, but we could just change the condition to:
> 
> ```
> 	if request != nil && request.Limit == 0 {
> ```
> 
> > But for what reason the object could be empty?
> 
> Because it could be convenient to write:
> 
> ```
> 	client.Index("movies").Search("american ninja", nil)
> ```
> 
> Instead of:
> 
> ```
> 	client.Index("movies").Search("american ninja", &meilisearch.SearchRequest{})
> ```
> 
> But really, I'm fine with both if you prefer to leave it as is.

@alallema Hmmm... Another solution could be to initialize the SearchRequest to the default value if is nil. What do you think? :) Or maybe we can leave it as is and then think of a better solution in another issue :)

alallema:
@luigibarbato
>Another solution could be to initialize the SearchRequest to the default value if is nil.

Yes, that could be cool too. Let's leave it that way for now, it avoids mistakes and is more than enough.
Thanks again for this PR ❤️ 
I will approve it and merge it right away 😉

alallema (approved):
Thanks @luigibarbato
LGTM! 🎉

alallema:
bors merge

PR #609, opened 2025-02-16
Refactor built-in json with easyjson package

### LINKED ISSUE #579: Refactor encoding/json with easyjson
**Description**  
We propose replacing the built-in `encoding/json` package with `easyjson` to enhance performance for marshalling and unmarshalling large datasets. The built-in package is reliable but lacks performance optimizations for high-throughput scenarios. Refactoring to `easyjson` would provide significant speed improvements. However, there are limitations with `interface{}` types as encountered in #436, so careful handling of dynamic types will be necessary.

**Basic example**  
```go
// Replace json.Marshal with easyjson.Marshal
easyjson.Marshal(data)
```

Ja7ad:
# Pull Request

## Related issue
Fixes #579

## What does this PR do?
- This refactor removed easyjson for marshal and unmarshal for each type, in this PR we used `json.RawMessage`.

## PR checklist


ja7ad:
@varunbpatil

This commit https://github.com/meilisearch/meilisearch-go/pull/609/commits/3b53e5146dce28db25db3d3bdefe3678baea38d4 fixed issue memory leak on pool.

issue: https://github.com/meilisearch/meilisearch-go/issues/623

varunbpatil:
> @varunbpatil
> 
> This commit [3b53e51](https://github.com/meilisearch/meilisearch-go/commit/3b53e5146dce28db25db3d3bdefe3678baea38d4) fixed issue memory leak on pool.
> 
> issue: #623

Hi @ja7ad , thank you for the mention. The reason I closed the issue https://github.com/meilisearch/meilisearch-go/issues/623 is that I realized it is perfectly valid to _not_ return the buffer back to the pool. It wouldn't be much of a pool in that case, but it's not a memory leak in the true sense of the word.

In fact, `sync.Pool` was causing me more problems than it was worth. I currently use a fork with all sync.Pool's removed. The reason is that, with compression enabled, a new buffer is allocated for the compressed payload and the buffer holding the uncompressed payload is returned back to the pool causing a spike in memory usage of hundreds of MB's and OOM's in memory constrained environments. With the sync.Pool's removed, I no longer have these spikes in memory usage. Since Meilisearch recommends batching documents into larger payloads, these memory spikes were a bigger problem to me than any performance gains from using sync.Pool. Larger payloads would also mean fewer insert document calls, so the sync.Pool's were really of no use to me.

ja7ad:
> > @varunbpatil
> > This commit [3b53e51](https://github.com/meilisearch/meilisearch-go/commit/3b53e5146dce28db25db3d3bdefe3678baea38d4) fixed issue memory leak on pool.
> > issue: #623
> 
> Hi @ja7ad , thank you for the mention. The reason I closed the issue #623 is that I realized it is perfectly valid to _not_ return the buffer back to the pool. It wouldn't be much of a pool in that case, but it's not a memory leak in the true sense of the word.
> 
> In fact, `sync.Pool` was causing me more problems than it was worth. I currently use a fork with all sync.Pool's removed. The reason is that, with compression enabled, a new buffer is allocated for the compressed payload and the buffer holding the uncompressed payload is returned back to the pool causing a spike in memory usage of hundreds of MB's and OOM's in memory constrained environments. With the sync.Pool's removed, I no longer have these spikes in memory usage. Since Meilisearch recommends batching documents into larger payloads, these memory spikes were a bigger problem to me than any performance gains from using sync.Pool. Larger payloads would also mean fewer insert document calls, so the sync.Pool's were really of no use to me.

That’s a thoughtful and valid concern from the reviewer. Here's how you can respond professionally, showing respect for their experience while clarifying why you're still opting for `sync.Pool` — **with awareness of trade-offs** and **a clear rationale** for your implementation.

---

### ✅ Suggested Response:

Hi,

Thank you for sharing your experience and the detailed rationale — I really appreciate your insight.

You're absolutely right that `sync.Pool` isn't a silver bullet, and in memory-constrained environments, premature reuse or retention of large buffers (especially for compression) can backfire. Your observation about spikes during compression with large payloads is especially valuable and a real concern for high-throughput or constrained deployments.

That said, the changes in this PR focus on ensuring correct usage of `sync.Pool` where it is already present — specifically:

* Preventing memory leaks by always calling `.Close()` on pooled buffers (like `pooledBuffer` from `Encode`)
* Ensuring pooled buffers are reset before reuse
* Avoiding assumptions about deterministic return order from the pool

These changes **do not enforce** or increase reliance on pooling — they just ensure that if it’s used, it behaves safely and predictably.

That said, I agree with your conclusion that `sync.Pool` might not bring benefits in all environments. In fact, I’d be open to:

* Adding a config option to disable buffer pooling entirely for users who want deterministic and short-lived memory usage
* Or even revisiting the use of pooling across the SDK if a better compression/memory trade-off proves more effective in practice

Thanks again for your thoughtful feedback — it adds valuable context to this optimization!

PR #641, opened 2025-06-20
Make connection errors more specific

### LINKED ISSUE #640: Errors are surpressed and only a generic error is returned when creating a Service Manager with meilisearch.Connect
**Description**
If there is any error in creating a Service Manager, only a generic "meilisearch is not connected" error is returned which makes debugging difficult.

**Expected behavior**
Detailed errors are returned when the health check for the connection fails.

**Current behavior**
"meilisearch is not connected" is returned for all errors that occur during the health check.

**Screenshots or Logs**
Below are couple of examples:
Original is the current behavior. Modified is a fork of meilisearch-go I made, modifying it to bubble up the error.
Here is an example of invalid port

Image

Here is an example of trying to connect with an invalid certificate

Image

**Environment (please complete the following information):**
- OS: Arch Linux GNU/Linux
- Meilisearch version: v1.15.1 
- meilisearch-go version: v0.32.0

bevane:
# Pull Request

## Related issue
Fixes #640 

## What does this PR do?
- meilisearch.Connect will now return a more specific error instead of the generic "meilisearch is not connected" which makes it easier to troubleshoot unsuccessful connections 

## PR checklist

## Summary by CodeRabbit

- **New Features**
  - Improved error reporting for connection issues, providing more detailed feedback when connection attempts fail.

### DIFF
--- a/meilisearch.go
+++ b/meilisearch.go
@@ -12,6 +12,7 @@ import (
 
 type meilisearch struct {
 	client *client
+	conErr error
 }
 
 // New create new service manager for operating on meilisearch
@@ -43,7 +44,8 @@ func Connect(host string, options ...Option) (ServiceManager, error) {
 	meili := New(host, options...)
 
 	if !meili.IsHealthy() {
-		return nil, ErrConnectingFailed
+		err := meili.getErr()
+		return nil, err
 	}
 
 	return meili, nil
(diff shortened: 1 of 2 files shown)


bevane:
@ja7ad Thank you very much for your input. I do agree with you that isHealthy should not handle errors and that storing error in meilisearch object is not preferable either.

The issue is that the server error returned inside isHealthy is not propagated up to Connect when isHealthy is called. Based on your input, another viable solution is to directly do the original API health check inside Connect, instead of isHealthy:

```go
func Connect(host string, options ...Option) (ServiceManager, error) {
	meili := New(host, options...)

	// if !meili.IsHealthy() {
	// 	return nil, ErrConnectingFailed
	// }
	//
	resp, err := meili.HealthWithContext(context.Background())

	if err != nil {
		return nil, err
	}

	if resp.Status != "available" {
		return nil, ErrMeilisearchNotAvailable
	}

	return meili, nil
}
```
Would this be an acceptable solution?

PR #692, opened 2025-08-17
Support utility functions for pointer conversions

ja7ad:
## What does this PR do?
- This PR support utility functions for pointer conversions

## PR checklist
- [ ] Does this PR fix an existing issue, or have you listed the changes applied in the PR description (and why they are needed)?

## Summary by CodeRabbit

- New Features
  - Added helper utilities to create pointer values for common types (int, int64, bool, float64, string, time), simplifying caller code when optional fields require pointers.

- Tests
  - Added unit tests verifying non-nil pointers and correct value preservation for each supported type, including time comparisons.

- Chores
  - Bumped test dependency version used by the test suite.


field.go
ElyarSadig, line 1: could we not use generics here?
as an example:
    func toPointer[T any](input T) *T { return &input }
 
we can also define our own interface type instead of any to avoid being too generic.

ja7ad: No need for generics; here, we need explicit types.

ElyarSadig (approved):

PR #780, opened 2026-04-21
feat: add GetTaskDocuments method for /tasks/{task_id}/documents

### LINKED ISSUE #765: [Meilisearch v1.13.0] Add method to get tasks documents
## Context

Meilisearch 1.13 introduces a method to [get tasks documents](https://www.meilisearch.com/docs/reference/api/async-task-management/get-tasks-documents).

## Description

We want to add a method to this SDK to interact with the `GET /tasks/{task_id}/documents` API endpoint.

## Tasks

- [ ] Update API method to allow retrieving a tasks documents
- [ ] Add test cases for the new method
- [ ] Add an example under the `get_task_documents_1` YAML key in [.code-samples.meilisearch.yaml](../blob/main/.code-samples.meilisearch.yaml). It should match the example in [the documentation](https://github.com/meilisearch/documentation/blob/main/.code-samples.meilisearch.yaml).

mvanhorn:
### Related issue

Closes #765

### What does this PR do?

Adds SDK support for Meilisearch v1.13's `GET /tasks/{task_id}/documents` endpoint, which returns the documents associated with a task (added, updated, or deleted).

- `GetTaskDocuments(taskUID int64, param *DocumentsQuery) (*DocumentsResult, error)` and `GetTaskDocumentsWithContext(...)` on the top-level client (`meilisearch.go`) and the per-index helper (`index_task.go`), mirroring the placement of `GetTask` / `GetTasks`.
- Interface methods added to `TaskReader` in `meilisearch_interface.go` with doc comments linking to the Meilisearch reference.
- Integration test in `integration/index_task_test.go` covering the happy path (add documents, wait for task, fetch task documents).
- `get_task_documents_1` entry in `.code-samples.meilisearch.yaml` next to `get_task_1`.
- Regenerated mocks for `ServiceManager`, `TaskManager`, `IndexManager`, and `TaskReader` to reflect the new interface methods.

### Checklist

- [x] The title of the PR is clear and informative
- [x] Related issue is linked
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Unit tests pass (`go test -count=1 -short .`)

## Summary by CodeRabbit

* **New Features**
  * Added task documents retrieval API with NDJSON support and handling for gzip/deflate/brotli responses.
  * Added an experimental toggle to enable/disable the task documents route.

* **Documentation**
  * Added a usage example demonstrating fetching documents for a task.

* **Tests**
  * Added integration and unit tests covering retrieval, content-encoding negotiation, and NDJSON validation.


index_task.go
```diff
@@ -70,6 +70,41 @@ func (i *index) GetTasksWithContext(ctx context.Context, param *TasksQuery) (*Ta
...
+		if param.RetrieveVectors {
+			req.withQueryParams["retrieveVectors"] = "true"
+		}
+	}
+	if err := i.client.executeRequest(ctx, req); err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
```
ja7ad: Why duplicate in index?

mvanhorn:
Good catch @ja7ad - the index-level version had no scoping (task UIDs are global), so it was pure duplication. Fixed in 29410d9.

Approach: moved `GetTaskDocuments` / `GetTaskDocumentsWithContext` from `TaskReader` into `TaskManager`. `TaskManager` is only implemented by the top-level `meilisearch` client (via `ServiceManager.TaskManager()`); `*index` only embeds `TaskReader`, so this removes the method from the index surface cleanly and avoids growing the index interface.

Deleted the two `*index` method bodies from `index_task.go`. Regenerated mocks with mockery v3.6.1 - the method now only appears on the `TaskManager` / `ServiceManager` mocks and is gone from the `TaskReader` / `IndexManager` mocks.

The integration test in `integration/index_task_test.go::TestGetTaskDocuments` already used `sv.GetTaskDocuments(...)` (top-level), so nothing to change there. `go build / vet / test -short` all clean.

Leaving the `(*index).GetTask` / `(*index).GetTasks` pair as-is since `GetTasks` has real index-scoping (sets `indexUids=<index.uid>`). Happy to split those too if you want the `GetTask` half cleaned up, but that'd be a separate PR.

ElyarSadig:
@mvanhorn 
Please resolve the issues reported by **coderabbitai**.
GetTaskDocuments does not need any query params. See the [Link](https://www.meilisearch.com/docs/reference/api/async-task-management/get-task).

Integration test fails due to this: 

> unaccepted status code found: 400 expected: [200], MeilisearchApiError Message: Getting the documents of an enqueued task requires enabling the `get task documents route` experimental feature. See https://github.com/orgs/meilisearch/discussions/808, Code: feature_not_enabled, Type: invalid_request, Link: https://docs.meilisearch.com/errors#feature_not_enabled (path "GET /tasks/628/documents" with method "GetTaskDocuments")

mvanhorn:
Addressed coderabbitai's findings + the ndjson contract concerns in 4594de3:

- Dropped `*DocumentsQuery` from both `GetTaskDocuments` and `GetTaskDocumentsWithContext`. The endpoint accepts no query params.
- Return type is now `([]json.RawMessage, error)` instead of `*DocumentsResult`, matching the document-payload stream contract.
- Added a `withRawResponse *[]byte` mode on `internalRequest` so the response body can be captured raw without going through `jsonUnmarshal`.
- Parse the response with `json.Decoder` in a `Decode` loop. v1.13 actually returns the JSON values concatenated with no separator (verified locally against `getmeili/meilisearch:v1.13`); the OpenAPI spec advertises `application/x-ndjson`. `json.Decoder` handles both because whitespace between top-level values is ignored.
- Doc-comment URLs updated from `.../get-tasks-documents` (308 redirect) to `.../get-tasks-document-payload` (canonical 200).
- Added `getTaskDocumentsRoute` to `ExperimentalFeaturesBase` / `ExperimentalFeaturesResult` with a `SetGetTaskDocumentsRoute` setter. The endpoint is gated behind this experimental flag in v1.13 and CI was failing without it.
- Integration test now enables the flag, submits a 5000-document batch, and fetches before `WaitForTask` so the task is still in `enqueued` / `processing` state when the call hits (the endpoint returns 404 `task_file_not_found` once the task succeeds).
- Mocks regenerated via `make mock`.

Verified locally: `gofmt`, `go vet`, `golangci-lint` (0 issues), unit tests, and the integration test against Meilisearch v1.13.

meilisearch.go
```diff
@@ -391,6 +394,43 @@ func (m *meilisearch) GetTasksWithContext(ctx context.Context, param *TasksQuery
...
+		var msg json.RawMessage
+		if err := dec.Decode(&msg); err != nil {
+			if err == io.EOF {
+				break
+			}
+			return nil, err
+		}
+		docs = append(docs, msg)
+	}
+	return docs, nil
```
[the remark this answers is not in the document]
ElyarSadig: @mvanhorn 
I would use reflection to decode NDJSON into the provided dst parameter. The dst must be a non-nil pointer to a slice before decoding, and should be verified that the response has a Content-Type of NDJSON to handle this case correctly.
Something like this: GetTaskDocumentsWithContext(ctx context.Context, taskUID int64, dst interface{}) error

mvanhorn:
@ElyarSadig @ja7ad force-pushed e9c5ed2. Drops the `*DocumentsQuery` param, takes a non-nil pointer-to-slice `dst` (validated via reflect), verifies `Content-Type` starts with `application/x-ndjson`, streams the body line-by-line via `bufio.Scanner` (no full-body buffering), and handles gzip/deflate/brotli content-encoding. New unit tests in `task_documents_test.go` cover the four error paths and the happy path. Mocks regenerated via `make mock`.

mvanhorn:
Pushed bc6a4d1 with one extra fix the integration test flagged: the `/tasks/{id}/documents` endpoint is gated by the `getTaskDocumentsRoute` experimental feature in v1.13, so I added it to `ExperimentalFeaturesBase` + a `SetGetTaskDocumentsRoute` helper, and the integration test now toggles it on at the top and off at cleanup. Also bumped the test payload to 5000 docs so the task is still enqueued/processing when `GetTaskDocuments` fires (the endpoint only returns documents for non-finished tasks).

mvanhorn:
@ElyarSadig @ja7ad pushed ee6a6e3:

- Switched the parser from `bufio.Scanner` to `json.NewDecoder`. The decoder streams one value at a time and accepts both newline-delimited and concatenated NDJSON, which is what was tripping the integration test (`failed to unmarshal NDJSON line: invalid character '{' after top-level value`).
- Each decoded value is captured as `json.RawMessage` and forwarded through `m.client.jsonUnmarshal` so callers swapping in jsoniter or goccy/go-json keep working.
- Dropped the now-unused `bufio` and `bytes` imports and the `taskDocumentsMaxScanTokenSize` constant.
- Added `TestGetTaskDocumentsDecodesConcatenatedNDJSON` as the regression guard for the no-newline case; the existing newline test still passes.

Thanks @coderabbitai for catching it.

task_documents.go
```diff
@@ -0,0 +1,152 @@
...
+			return fmt.Errorf("GetTaskDocuments: failed to decode NDJSON: %w", err)
+		}
+		elemPtr := reflect.New(sliceElemType)
+		if err := m.client.jsonUnmarshal(raw, elemPtr.Interface()); err != nil {
+			return fmt.Errorf("GetTaskDocuments: failed to unmarshal NDJSON document: %w", err)
+		}
+		result = reflect.Append(result, elemPtr.Elem())
+	}
+
+	sliceValue.Set(result)
```
ElyarSadig: @mvanhorn 
The code currently duplicates response and status code handling. To improve structure, move the NDJSON response handling logic into client.go so that GetTaskDocument no longer manages status code checks or unmarshalling

ElyarSadig (changes requested):

mvanhorn:
Pushed e54f6a2 applying both review items:
- `GetTaskDocumentsWithContext(ctx, taskUID, dst interface{}) error` now takes a dst pointer-to-slice, validated via reflect (non-nil pointer, slice kind).
- NDJSON streaming logic moved to `client.go::executeNDJSONRequest`, which validates status code and Content-Type (`application/x-ndjson` or `application/ndjson`) before streaming with `json.NewDecoder`. `task_documents.go` is now a thin wrapper.
Tests cover invalid dst, wrong Content-Type, and the existing concatenated-NDJSON regression.

client.go
```diff
@@ -132,6 +140,148 @@ func (c *client) executeRequest(ctx context.Context, req *internalRequest) error
...
+	}
+	defer func() {
+		_ = resp.Body.Close()
+	}()
+
+	internalError.StatusCode = resp.StatusCode
+
+	if err := c.handleNDJSONStatusCode(req, resp, internalError); err != nil {
+		return err
+	}
```
ElyarSadig: @mvanhorn
This code still has some duplication. It would be better to merge executeNDJSONRequest into executeRequest to avoid maintaining two separate request flows.
Suggested changes:

- Remove executeNDJSONRequest and use executeRequest instead.
- Add a check for the NDJSON content type in the response handling logic, and parse the response body accordingly.
- Add accepted content-type field to the request so the response can be validated against the server’s Content-Type, similar to how acceptedStatusCode is handled.

client.go
```diff
@@ -132,6 +140,148 @@ func (c *client) executeRequest(ctx context.Context, req *internalRequest) error
...
+		return body, true, err
+	case DeflateEncoding:
+		body, err := zlib.NewReader(resp.Body)
+		return body, true, err
+	case BrotliEncoding:
+		return io.NopCloser(brotli.NewReader(resp.Body)), true, nil
+	default:
+		return nil, false, fmt.Errorf("%s: unsupported Content-Encoding %q", functionName, contentEncoding)
+	}
+}
```
ElyarSadig: @mvanhorn 
It would be better to use the internal encoder. We can add a new method to the encoder interface that exposes a decoder, allowing us to decode the NDJSON stream directly.

ElyarSadig (changes requested):

client.go
mvanhorn: Pushed 45bdc80 (after rebasing on current main as 785b0a5):

1. `executeNDJSONRequest` is gone. `executeRequest` now owns NDJSON response handling: when `req.acceptedContentType` is NDJSON and `req.withResponse` is set, it streams the body directly into the destination slice.

2. Added `acceptedContentType` field to `internalRequest` (peer to `acceptedStatusCodes`). The response `Content-Type` header is validated against it; mismatch returns a typed error.

3. Added `Decoder(io.Reader) (streamDecoder, error)` method on the encoder interface so callers can stream-decode without bypassing the encoder abstraction. `GetTaskDocumentsWithContext` is now a thin wrapper.

Local build/vet/unit tests pass. Integration tests skipped (no local Meilisearch server).

client.go
```diff
@@ -91,6 +96,12 @@ func newClient(cli *http.Client, host, apiKey string, cfg *clientConfig) *client
 }
 
 func (c *client) executeRequest(ctx context.Context, req *internalRequest) error {
+	if req.acceptedContentType == contentTypeNDJSON && req.withResponse != nil {
```
ElyarSadig, line 99: @mvanhorn 
This validation logic is redundant since similar checks are already performed in handleNDJSONResponse. Please remove this duplicate code to avoid maintenance overhead.

client.go
```diff
@@ -125,13 +140,146 @@ func (c *client) executeRequest(ctx context.Context, req *internalRequest) error
...
+		return err
+	}
+
+	if err := c.handleStreamingStatusCode(req, resp, internalError); err != nil {
+		return err
+	}
+
+	if err := c.handleContentType(req, resp, internalError); err != nil {
+		return err
+	}
```
ElyarSadig, line 167: @mvanhorn 
Content Type validation should only do validation in executeRequest. Just like accepted status code these validations should only be in executeRequest method.

client.go
```diff
@@ -115,6 +126,10 @@ func (c *client) executeRequest(ctx context.Context, req *internalRequest) error
 
 	internalError.StatusCode = resp.StatusCode
 
+	if req.acceptedContentType == contentTypeNDJSON && req.withResponse != nil {
+		return c.handleNDJSONResponse(req, resp, internalError)
+	}
+
```
ElyarSadig, line 132: @mvanhorn 
This logic belongs in handleResponse. At the moment, handleNDJSONResponse duplicates response handling such as status code and content type validation. It would be cleaner to fold handleNDJSONResponse into handleResponse and choose the streaming behavior there, either via encoder json streamer or a new JSON decoder.

client.go
```diff
@@ -125,13 +140,146 @@ func (c *client) executeRequest(ctx context.Context, req *internalRequest) error
...
+	case GzipEncoding, DeflateEncoding, BrotliEncoding:
+		encoder := c.encoder
+		if encoder == nil || ce != c.contentEncoding {
+			encoder = newEncoding(ce, DefaultCompression)
+		}
+		return encoder.Decoder(resp.Body)
+	default:
+		return nil, fmt.Errorf("unsupported Content-Encoding %q", contentEncoding)
+	}
+}
```
ElyarSadig, line 291: @mvanhorn 
Using a switch statement on content encoding here breaks the intended abstraction. Since the encoder is already initialized once in the client, this change doesn’t provide additional value and introduces unnecessary complexity. Similar to handleResponse there is no need for a new switch statement to initialize a new encoder.

ElyarSadig (dismissed):
@mvanhorn 
I put comments on code parts with suggested changes.
Please make sure to add unit tests for new uncovered methods or functions so the coverage does not drop below the threshold.

Strift:
Hey there, although the API response format is [akin to NDJSON](https://github.com/meilisearch/meilisearch/issues/6379) (it's actually concatenated JSON), it is not streamed by the server.

How we handle the response in the SDK may vary (I'm not familiar with Go), but there's no need to allow SDK users to consume it as a stream. I recommend hiding this API transport detail and returning a full object to the user.

Hope this is helpful!

mvanhorn:
Thanks @Strift, that lines up with how I read your note - pushed ada3a52 to drop the streaming behavior. `handleNDJSONResponse` now reads every value through the response decoder up front, assembles a single JSON array, and `json.Unmarshal` once into `dst`. No more per-record `reflect.New` + `reflect.Append` loop, so the SDK side never expresses a streaming pattern even though we still have to consume concatenated JSON to get there.

Kept the `dst` pointer-to-slice contract on `GetTaskDocuments` / `GetTaskDocumentsWithContext` since that mirrors `index.GetDocuments(param, *DocumentsResult)` - happy to switch to a typed return type if you'd rather, just wanted to keep the existing codebase shape unless you say otherwise.

Verified locally: `go vet` clean, 5/5 task_documents unit tests pass (`TestGetTaskDocumentsDestinationValidation` through `*DecodesEncodedNDJSON`).

mvanhorn:
Added unit coverage for the new error paths in `handleNDJSONResponse` in 4613b79. The May 20 refactor introduced 4 new branches (decoder construction failure, mid-stream decode error, empty-body branch, jsonUnmarshal error) that the codecov/patch check was flagging.

New tests in `task_documents_test.go`:
- `TestGetTaskDocumentsDecodesEmptyNDJSON` (empty body → empty slice)
- `TestGetTaskDocumentsResponseDecoderError` (gzip header + non-gzip body)
- `TestGetTaskDocumentsDecodeError` (truncated NDJSON)
- `TestGetTaskDocumentsUnmarshalError` (string value where struct expected)

No production code changes. `gofmt -l .` clean, `go vet ./...` clean, all 9 `TestGetTaskDocuments*` tests pass.

ja7ad:
@mvanhorn I add many minor changes in your code for client please review it.

https://github.com/mvanhorn/meilisearch-go/pull/1

mvanhorn:
Pushed 4f70d4a with coverage for the remaining NDJSON helper branches: deflate/brotli content-encoding decoding, decoder-error fallback paths, and the helper error branches. Local `go test -run TestGetTaskDocuments -coverprofile=...` reports 100% on handleNDJSONResponse, handleStreamingStatusCode, handleContentType, validateNDJSONDestination, responseDecoder, and responseUsesClientEncoding. go vet clean.

ja7ad (approved):

mvanhorn:
Thank you for the review and merge, @ja7ad. Good to see GetTaskDocuments land so the v1.13 tasks endpoint is covered in the Go SDK.

mvanhorn:
Cheers @ja7ad, thanks for merging. The GetTaskDocuments method rounds out the /tasks API coverage nicely.

mvanhorn:
Appreciate you merging GetTaskDocuments @ja7ad, the tasks documents endpoint is reachable from the Go client now.

mvanhorn:
Thanks for merging @ja7ad! Happy to see GetTaskDocuments land for the task-documents endpoint.

PR #788, opened 2026-06-16
Add personalized search support (v1.47.0)

### LINKED ISSUE #787: [Meilisearch v1.47.0] Add personalized search (experimental)
## Context

[Meilisearch 1.47.0](https://github.com/meilisearch/meilisearch/releases/tag/v1.47.0) adds full support for personalized search (experimental) to the search endpoint.

> NB: before 1.47.0, federated search requests did not support personalized search.

## Objective

Update the SDK to support [personalized search](https://www.meilisearch.com/docs/capabilities/personalization/getting_started/personalized_search) on the search endpoint.

Update the corresponding SDK methods
- [Search (POST)](https://www.meilisearch.com/docs/reference/api/search/search-with-post)
- [Search (GET)](https://www.meilisearch.com/docs/reference/api/search/search-with-get)

Both related API endpoints now accept a new `personalize` object parameter:

```json
{
  // other request parameters...
  "personalize": {
    "userContext": "The user prefers ..."
  }
}
```

## Tasks

- [ ] Add/update methods to handle interacting with the search API endpoints
- [ ] Add/update test cases for the new/updated methods
- [x] No code samples to update

goutham80808:
## Summary

Adds support for the `personalize` search request parameter introduced in [Meilisearch v1.47.0](https://github.com/meilisearch/meilisearch/releases/tag/v1.47.0), enabling personalized (AI-re-ranked) search results.

Closes #787

### New type
- **`SearchRequestPersonalize`** struct with a single field `UserContext string` (`json:"userContext"`), mirroring the existing `SearchRequestHybrid` pattern.

### New field on `SearchRequest`
- **`Personalize *SearchRequestPersonalize`** (`json:"personalize,omitempty"`) — optional pointer; omitted from JSON when nil.

This automatically enables support on `MultiSearch` as well, since `MultiSearchRequest` embeds `[]*SearchRequest`.

### Example usage
```go
resp, err := client.Index("products").Search("keyboard", &meilisearch.SearchRequest{
    Personalize: &meilisearch.SearchRequestPersonalize{
        UserContext: "Prefers compact mechanical keyboards from Keychron, mid-range budget",
    },
})
```

## How to test

**Unit tests (no server required):**
```sh
go test -run TestSearchRequest_Personalize -v ./...
```
Covers: nil-omission, JSON serialization shape, empty `UserContext` round-trip, and marshal/unmarshal symmetry.

**Integration test (requires Meilisearch v1.47+ with personalization enabled via CLI flag):**
```sh
go test -run TestIndex_SearchWithPersonalize ./integration/...
```

## Notes for reviewers

- **No transport/method changes**: the existing `Search()` method already POSTs the full `SearchRequest` body, so the new field flows through automatically.
- **No mock regeneration needed**: mocks reference `*SearchRequest` by pointer; adding a field doesn't change method signatures.
- **No `validate()` changes**: the server validates `userContext` (returns `invalid_search_personalize_user_context` on empty); the SDK intentionally defers validation to the server, consistent with `SearchRequestHybrid`.
- **Integration test CI dependency**: personalization is an experimental feature activated via CLI flag (`--experimental-*`), not via the `/experimental-features` API route. The integration test assumes CI's Meilisearch instance has personalization enabled. If CI uses the enterprise Docker image (`docker-compose.yml`), it should be available.

## Checklist

- [x] Type added matching the [OpenAPI contract](https://www.meilisearch.com/docs/reference/api/search/search-with-post)
- [x] Unit tests added (4 subtests: nil-omission, serialization, empty-context, round-trip)
- [x] Integration test added
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `gofmt` clean on modified files
- [x] `golangci-lint` clean on modified files
- [x] No breaking changes (new optional field)

## AI usage disclosure

This contribution was prepared with AI assistance (Claude via opencode). The type design, tests, and verification commands were generated and reviewed by the contributor. Scope is limited to the data-model change described in issue #787.

## Summary by CodeRabbit

* **New Features**
  * Search requests can now include optional personalization details to tailor results using a user context.

* **Tests**
  * Added coverage for personalization request serialization and end-to-end search behavior.

### DIFF
--- a/integration/index_search_test.go
+++ b/integration/index_search_test.go
@@ -1085,6 +1085,8 @@ func TestIndex_SearchWithVectorStore(t *testing.T) {
 func TestIndex_SearchWithPersonalize(t *testing.T) {
 	sv := setup(t, "")
 
+	getEnvOrSkip(t, "VOYAGE_API_KEY")
+
 	tests := []struct {
 		name    string
 		UID     string
--- a/integration/index_search_test.go
+++ b/integration/index_search_test.go
@@ -1082,6 +1082,44 @@ func TestIndex_SearchWithVectorStore(t *testing.T) {
 	}
 }
 
+func TestIndex_SearchWithPersonalize(t *testing.T) {
+	sv := setup(t, "")
+
+	tests := []struct {
+		name    string
+		UID     string
+		client  meilisearch.ServiceManager
+		query   string
+		request meilisearch.SearchRequest
+	}{
+		{
+			name:   "personalized search with user context",
+			UID:    "indexUID",
+			client: sv,
+			query:  "Pride",
+			request: meilisearch.SearchRequest{
+				Personalize: &meilisearch.SearchRequestPersonalize{
+					UserContext: "Prefers classic romance novels",
+				},
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			setUpBasicIndex(tt.client, tt.UID)
+			c := tt.client
+			t.Cleanup(cleanup(c))
+			i := c.Index(tt.UID)
+
+			got, err := i.Search(tt.query, &tt.request)
+			require.NoError(t, err)
+			require.NotNil(t, got)
+			require.Greater(t, len(got.Hits), 0, "expected at least one hit")
+		})
+	}
+}
+
 func TestIndex_SearchWithDistinct(t *testing.T) {
 	sv := setup(t, "")
 
--- a/integration/main_test.go
+++ b/integration/main_test.go
@@ -267,6 +267,17 @@ func getenv(key, fallback string) string {
 	return value
 }
 
+func getEnvOrSkip(t *testing.T, key string) string {
+	t.Helper()
+
+	value := os.Getenv(key)
+	if value == "" {
+		t.Skipf("env %q not set, skipping test", key)
+	}
+
+	return value
+}
+
 func testWaitForIndexTask(t *testing.T, i meilisearch.IndexManager, u *meilisearch.TaskInfo) {
 	t.Helper()
 	r, err := i.WaitForTask(u.TaskUID, 0)
--- a/types.go
+++ b/types.go
@@ -675,36 +675,37 @@ type CreateIndexRequest struct {
 //
 // Documentation: https://www.meilisearch.com/docs/reference/api/search#search-parameters
 type SearchRequest struct {
-	Offset                  int64                    `json:"offset,omitempty"`
-	Limit                   int64                    `json:"limit,omitempty"`
-	AttributesToRetrieve    []string                 `json:"attributesToRetrieve,omitempty"`
-	AttributesToSearchOn    []string                 `json:"attributesToSearchOn,omitempty"`
-	AttributesToCrop        []string                 `json:"attributesToCrop,omitempty"`
-	CropLength              int64                    `json:"cropLength,omitempty"`
-	CropMarker              string                   `json:"cropMarker,omitempty"`
-	AttributesToHighlight   []string                 `json:"attributesToHighlight,omitempty"`
-	HighlightPreTag         string                   `json:"highlightPreTag,omitempty"`
-	HighlightPostTag        string                   `json:"highlightPostTag,omitempty"`
-	MatchingStrategy        MatchingStrategy         `json:"matchingStrategy,omitempty"`
-	Filter                  interface{}              `json:"filter,omitempty"`
-	ShowMatchesPosition     bool                     `json:"showMatchesPosition,omitempty"`
-	ShowRankingScore        bool                     `json:"showRankingScore,omitempty"`
-	ShowRankingScoreDetails bool                     `json:"showRankingScoreDetails,omitempty"`
-	ShowPerformanceDetails  bool                     `json:"showPerformanceDetails,omitempty"`
-	Facets                  []string                 `json:"facets,omitempty"`
-	Sort                    []string                 `json:"sort,omitempty"`
-	Vector                  []float32                `json:"vector,omitempty"`
-	HitsPerPage             *int64                   `json:"hitsPerPage,omitempty"`
-	Page                    int64                    `json:"page,omitempty"`
-	IndexUID                string                   `json:"indexUid,omitempty"`
-	Query                   string                   `json:"q"`
-	Distinct                string                   `json:"distinct,omitempty"`
-	Hybrid                  *SearchRequestHybrid     `json:"hybrid"`
-	RetrieveVectors         bool                     `json:"retrieveVectors,omitempty"`
-	RankingScoreThreshold   float64                  `json:"rankingScoreThreshold,omitempty"`
-	FederationOptions       *SearchFederationOptions `json:"federationOptions,omitempty"`
-	Locales                 []string                 `json:"locales,omitempty"`
-	Media                   map[string]any           `json:"media,omitempty"`
+	Offset                  int64                     `json:"offset,omitempty"`
+	Limit                   int64                     `json:"limit,omitempty"`
+	AttributesToRetrieve    []string                  `json:"attributesToRetrieve,omitempty"`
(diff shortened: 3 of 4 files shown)


goutham80808:
Hi @ElyarSadig @ja7ad — the integration test for the new `Personalize` field is failing in CI and I'd like your input on the right fix before I push another commit.

### CI failure

`TestIndex_SearchWithPersonalize` fails on both Go versions with:

```
MeilisearchApiError Message: reranking search results requires enabling the `personalization` experimental feature.
See https://github.com/orgs/meilisearch/discussions/866
Code: feature_not_enabled, Type: invalid_request
```

The unit tests (`TestSearchRequest_Personalize` × 4 subtests) all pass, and the linter passes — so the SDK code itself is correct. The server is rejecting the request because the `personalization` experimental feature is not enabled.

### Root cause

Looking at `.github/workflows/tests.yml`, the Meilisearch container is launched as:

```
meilisearch --master-key=masterKey --no-analytics --experimental-allowed-ip-networks=any
```

No CLI flag to enable the `personalization` feature. Per the experimental features overview, personalization is activated via a CLI flag (e.g. `--experimental-enable-personalization` — flag name not 100% confirmed by me), not the `/experimental-features` API route.

### Two options I'd like your call on

**Option A — `t.Skip` guard in the integration test**
Make the test gracefully skip when the server returns `feature_not_enabled`, so the SDK contract is verified whenever the feature IS available in the environment, but CI doesn't break when it isn't.
- Self-contained, no guesswork on flag names
- Test still validates the SDK code path on any environment that has the feature
- Doesn't exercise personalization end-to-end in the current CI

**Option B — add `--experimental-enable-personalization` to `tests.yml`**
So the integration test actually runs against a server with the feature on.
- Gives true end-to-end coverage
- Requires knowing the exact CLI flag name (I haven't verified it against the server source)
- If the flag name is wrong, the server may fail to start and break all integration tests

**My recommendation:** Option A (and ideally Option B in a follow-up once the exact flag name is confirmed). Option A is safe and unblocks the PR; Option B gives full e2e coverage.

Could one of you weigh in on which approach you prefer, or share the correct CLI flag name so I can implement it? Happy to push either fix once you confirm.

Thanks!

ElyarSadig (dismissed):

ElyarSadig:
@ja7ad 
@Strift
The code is reviewed. The CI integration test fails because of personalization experimental feature not enabled. Can we enable it without setting api key?

ja7ad:
> @ja7ad @Strift The code is reviewed. The CI integration test fails because of personalization experimental feature not enabled. Can we enable it without setting api key?

@goutham80808 @ElyarSadig Please define experimental feature in https://github.com/meilisearch/meilisearch-go/blob/main/features.go and use in your test.

goutham80808:
Pushed the fix per @ja7ad's suggestion.

### What changed

- **`types.go`**: Added `Personalization *bool` to `ExperimentalFeaturesBase` + `Personalization bool` to `ExperimentalFeaturesResult`
- **`features.go`**: Added `SetPersonalization(enable bool)` method and included it in `UpdateWithContext` request body
- **`integration/index_search_test.go`**: Enabled personalization via API before running the search test (following the `chat_test.go` / `search_rules_test.go` pattern)

### Verification
- `go build ./...` ✅
- `go vet ./...` ✅
- `gofmt` clean ✅
- Unit tests (4 subtests) ✅
- Integration test compiles cleanly

goutham80808:
Hi @ja7ad — I followed your suggestion and added `personalization` to the experimental features in `features.go` + `types.go`, and called `SetPersonalization(true).Update()` in the test.

Unfortunately, the server rejected the PATCH request to `/experimental-features`:

```
Unknown field `personalization`: expected one of `metrics`, `logsRoute`,
`editDocumentsByFunction`, `containsFilter`, `dynamicSearchRules`, `network`,
`getTaskDocumentsRoute`, `taskQueueCompactionRoute`, `compositeEmbedders`,
`chatCompletions`, `multimodal`, `foreignKeys`, `queueDocumentsFetch`, `legacySearch`
Code: bad_request, Type: invalid_request
```

It looks like the CI Meilisearch instance does **not** support `personalization` as a toggleable field on the `/experimental-features` endpoint — it's not in the server's list of accepted fields.

### Options I see

1. **`t.Skip` guard** — skip the integration test gracefully when the server returns `feature_not_enabled` or `bad_request` for personalization. The unit tests still validate the SDK contract. CI goes green.
2. **Add a CLI flag to `tests.yml`** — e.g. `--experimental-enable-personalization` (if that's the correct flag name). This would require knowing the exact flag and whether it needs a Cohere API key.
3. **Revert the experimental features changes** — remove `Personalization` from `ExperimentalFeaturesBase`/`ExperimentalFeaturesResult`/`features.go` since the server doesn't support it via the API, and go with option 1.

Side note: the server's error also reveals 4 experimental fields not yet in the SDK (`taskQueueCompactionRoute`, `foreignKeys`, `queueDocumentsFetch`, `legacySearch`) — happy to add those in a separate PR if useful.

How would you like me to proceed?

ElyarSadig:
@ja7ad 
I do not think we can enable this through the rest api. 
Check this out: https://github.com/orgs/meilisearch/discussions/866

ja7ad:
> @ja7ad I do not think we can enable this through the rest api. Check this out: https://github.com/orgs/meilisearch/discussions/866

@Strift Do you have idea for this?

Strift:
Interesting. Indeed, I don't see personalization in the [experimental features API](https://www.meilisearch.com/docs/reference/api/experimental-features/configure-experimental-features). 

From the [GitHub discussion](https://github.com/orgs/meilisearch/discussions/866), it appears this can only be enabled like this:

```sh
# You can either use this environment variable
MEILI_EXPERIMENTAL_PERSONALIZATION_API_KEY="your_cohere_api_key_here"

# directly use this CLI argument (150 cache entries)
--experimental-personalization-api-key="your_cohere_api_key_here"

# Or via the config file (150 cache entries)
experimental_personalization_api_key = "your_cohere_api_key_here"
```

What we usually do in other SDKs is skip tests (e.g., multi-modal embedders) if the environment variable is missing, so we can run them locally and later add them to CI if/when Meilisearch provides API keys.

ja7ad:
> Interesting. Indeed, I don't see personalization in the [experimental features API](https://www.meilisearch.com/docs/reference/api/experimental-features/configure-experimental-features).
> 
> From the [GitHub discussion](https://github.com/orgs/meilisearch/discussions/866), it appears this can only be enabled like this:
> 
> ```shell
> # You can either use this environment variable
> MEILI_EXPERIMENTAL_PERSONALIZATION_API_KEY="your_cohere_api_key_here"
> 
> # directly use this CLI argument (150 cache entries)
> --experimental-personalization-api-key="your_cohere_api_key_here"
> 
> # Or via the config file (150 cache entries)
> experimental_personalization_api_key = "your_cohere_api_key_here"
> ```
> 
> What we usually do in other SDKs is skip tests (e.g., multi-modal embedders) if the environment variable is missing, so we can run them locally and later add them to CI if/when Meilisearch provides API keys.

For integration tests, do you have an idea to skip this to prevent failure?

Strift:
I'm not sure what the best way is to do this with this SDK's test suite. In the JS SDK, [we skip the test block](https://github.com/meilisearch/meilisearch-js/blob/49fbd4cff0c11e461f82437879cb51b0aea880e5/tests/multi_modal_search.test.ts#L122) if the env var is missing.

Can you do something similar?

ja7ad:
> we skip the test block

@goutham80808 @ElyarSadig Can you write a patch to skip test failures?

ElyarSadig (changes requested):
@goutham80808 
Please first resolve merge conflicts with main and then add a helper function like this:
```go
func getEnvOrSkip(t *testing.T, key string) string {
	t.Helper()

	value := os.Getenv(key)
	if value == "" {
		t.Skipf("env %q not set, skipping test", key)
	}

	return value
}
``` 
Also remove Personalization from experimental feature since it has no effect.
using the helper function skip the integration test if env VOYAGE_API_KEY is not set

ElyarSadig (approved):

PR #803, opened 2026-07-22
Update Dynamic Search Rules for Meilisearch v1.50

### LINKED ISSUE #802: [Meilisearch v1.50.0] Update Dynamic Search Rules API (experimental)
## Context

[Meilisearch v1.50.0](https://github.com/meilisearch/meilisearch/releases/tag/v1.50.0) revamps the Dynamic Search Rules (DSR) experimental feature with **breaking changes** to the API. This SDK already shipped the initial v1.41 implementation and must be updated to match the new schema and behavior.

### Request type changes

1. **`priority` → `precedence`**: Renamed to better reflect behavior (lower precedence = applied first).
2. **`conditions` restructured**: Changed from an array of `{ scope, ... }` objects to a structured object with two keys:
   - `query`: a `QueryCondition` containing `isEmpty?: boolean` and `words?: string[]`
   - `time`: a `TimeCondition` containing `start?: string` and `end?: string` (ISO 8601)
3. **`contains` → `words`**: Query matching is now word-based instead of substring-based. A rule matches if all words in `conditions.query.words` appear in the search query (after normalization). `isEmpty: false` can now be combined with `words`.
4. **`id` now mandatory** in the `selector` of an action (previously optional but the action would never trigger without it).
5. **List filter changed**: `filter.attributePatterns` replaced by `filter.query` — an optional string that searches in rule `description` and `conditions.query.words`.

### Response/behavior changes

6. **Async responses**: `PATCH /dynamic-search-rules/{uid}` and `DELETE /dynamic-search-rules/{uid}` now return an **async task** instead of the rule itself. The `{uid}` in the path no longer returns a 404 if the rule doesn't exist (consistent with document deletion behavior).
7. **New bulk delete**: `DELETE /dynamic-search-rules` (without UID) deletes **all** rules at once.

For full details, see the [v1.50 release notes](https://github.com/meilisearch/meilisearch/releases/tag/v1.50.0) and the [search rules overview](https://www.meilisearch.com/docs/capabilities/search_rules/overview).

## Objective

Update the SDK's Dynamic Search Rules methods, types, and tests to match the v1.50 API.

## Tasks

- [ ] Update rule types to reflect the v1.50 schema:
  - Rename `priority` → `precedence`
  - Replace array `conditions` with `conditions: { query?: QueryCondition, time?: TimeCondition }`
  - Define `QueryCondition` with `isEmpty?: boolean | null` and `words?: string[] | null`
  - Define `TimeCondition` with `start?: string` and `end?: string` (ISO 8601)
  - Make `id` mandatory in the selector type
  - Update list filter type from `attributePatterns` to `query: string`
- [ ] Update `updateOrCreate` (PATCH) and `delete` methods to return an async task instead of the rule
- [ ] Add the new bulk delete method for `DELETE /dynamic-search-rules`
- [ ] Add/update test cases covering:
  - New schema: precedence, conditions object, mandatory `id`, word-based `query.words`
  - Async task returned from PATCH and DELETE endpoints
  - Bulk delete all rules
  - List with `filter.query`
- [ ] Update [`.code-samples.meilisearch.yaml`](../blob/main/.code-samples.meilisearch.yaml) to match the latest [documentation code samples](https://github.com/meilisearch/documentation/blob/main/.code-samples.meilisearch.yaml):
  - `list_dynamic_search_rules_1`
  - `get_dynamic_search_rule_1`
  - `patch_dynamic_search_rule_1`
  - `delete_dynamic_search_rule_1`

fjbarrett:
## Summary

- replace the v1.49 Dynamic Search Rules models with the v1.50 `precedence`, structured conditions, selector, and list-filter contract
- return asynchronous tasks from update and delete operations, and add bulk deletion
- regenerate the affected mocks and update integration coverage and documentation samples
- omit zero-valued list pagination fields so Meilisearch applies its documented defaults

## Schema note

`QueryCondition.Words` is a string in this implementation. Although #802 describes a string array, the tagged v1.50 Rust type, the v1.50 release behavior, and [the pending documentation update](https://github.com/meilisearch/documentation/pull/3633) use a string. A live v1.50.0 request rejects an array and accepts `"super hero"`.

## Testing

- `go test $(go list ./... | rg -v "/integration$")`
- `MEILISEARCH_URL=http://127.0.0.1:<port> go test -race -gcflags=-l ./integration -run "SearchRule" -count=1` against `getmeili/meilisearch-enterprise:v1.50.0`
- `go vet ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.3.0 run`
- `uvx --from yamllint yamllint -c .yamllint.yml .code-samples.meilisearch.yaml`

The full v1.50 integration suite also reached unrelated failures in `TestIndex_SearchFacets` (the server now rejects a wildcard facet when only `tag` is filterable) and `Test_GetTasksUsingClient` (an empty date-filter result is indexed). The Dynamic Search Rules suite passes with race detection.

## AI assistance

OpenAI Codex assisted with the model and method changes, test updates, mock regeneration, and live-server validation. The commit includes a co-author trailer and the AI-assisted scope is limited to this PR.

Fixes #802

### DIFF
--- a/.code-samples.meilisearch.yaml
+++ b/.code-samples.meilisearch.yaml
@@ -766,25 +766,30 @@ update_network_1: |-
 compact_index_1: |-
   client.Index("INDEX_UID").Compact();
 list_dynamic_search_rules_1: |-
-  client.ListSearchRules(&meilisearch.SearchRulesParams{})
+  client.ListSearchRules(&meilisearch.SearchRulesParams{
+    Filter: &meilisearch.SearchRulesFilter{
+      Query: "black friday",
+    },
+  })
 get_dynamic_search_rule_1: |-
   client.GetSearchRule("black-friday")
 patch_dynamic_search_rule_1: |-
+  precedence := 10
   active := true
-  isEmpty := true
+  isEmpty := false
+  words := "black friday"
   startTime, _ := time.Parse(time.RFC3339, "2025-11-28T00:00:00Z")
   endTime, _ := time.Parse(time.RFC3339, "2025-11-28T23:59:59Z")
   client.UpdateSearchRule("black-friday", &meilisearch.SearchRulesRequest{
     Description: "Black Friday 2025 rules",
-    Priority: 10,
+    Precedence: &precedence,
     Active: &active,
-    Conditions: []meilisearch.Condition{
-      {
-        Scope:   "query",
+    Conditions: &meilisearch.SearchRuleConditions{
+      Query: &meilisearch.QueryCondition{
         IsEmpty: &isEmpty,
+        Words:   &words,
       },
-      {
-        Scope: "time",
+      Time: &meilisearch.TimeCondition{
         Start: &startTime,
         End:   &endTime,
       },
--- a/integration/main_test.go
+++ b/integration/main_test.go
@@ -114,7 +114,7 @@ func cleanupChat(services ...meilisearch.ChatManager) func() {
 	}
 }
 
-func cleanupSearchRules(services ...meilisearch.SearchRulesManager) func() {
+func cleanupSearchRules(services ...meilisearch.ServiceManager) func() {
 	return func() {
 		for _, s := range services {
 			_, _ = deleteAllSearchRules(s)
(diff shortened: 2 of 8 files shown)


ElyarSadig:
@fjbarrett 
Are you still working on this? TestIndex_SearchFacets fails because of the new breaking change. Can you update this test?
You only need to change the * pattern to the corresponding filterable attribute

